PR #3, opened 2019-07-02
Add examples of basic and digest authentication [skip ci]

razonyang:
| Q             | A
| ------------- | ---
| Is bugfix?    | 
| New feature?  |
| Breaks BC?    |
| Tests pass?   |
| Fixed issues  |

```
$ curl http://localhost:8080/basic-auth --user foo:bar
Hi foo

$ curl http://localhost:8080/digest-auth --user fzz:baz --digest
Hi fzz
```

### DIFF
--- a/src/Factory/AppRouterFactory.php
+++ b/src/Factory/AppRouterFactory.php
@@ -7,14 +7,23 @@ use Yiisoft\Router\Route;
 use Yiisoft\Router\RouterFactory;
 use Yiisoft\Yii\Web\Middleware\Controller;
 use App\Controller\SiteController;
+use Middlewares\BasicAuthentication;
+use Middlewares\DigestAuthentication;
+use App\Http\MiddlewareChain;
 
 class AppRouterFactory
 {
     public function __invoke(ContainerInterface $container)
     {
+        $basicAuth = $container->get(BasicAuthentication::class);
+        $digestAuth = $container->get(DigestAuthentication::class);
+        $authorizedAction = new Controller(SiteController::class, 'auth', $container);
+
         $routes = [
             Route::get('/')->to(new Controller(SiteController::class, 'index', $container)),
-            Route::get('/test/{id:\w+}')->to(new Controller(SiteController::class, 'testParameter', $container))
+            Route::get('/test/{id:\w+}')->to(new Controller(SiteController::class, 'testParameter', $container)),
+            Route::get('/basic-auth')->to(new MiddlewareChain($basicAuth, $authorizedAction)),
+            Route::get('/digest-auth')->to(new MiddlewareChain($digestAuth, $authorizedAction)),
         ];
 
         return (new RouterFactory(new FastRouteFactory(), $routes))($container);
--- a/src/Http/MiddlewareChain.php
+++ b/src/Http/MiddlewareChain.php
@@ -1,0 +1,79 @@
+<?php
+namespace App\Http;
+
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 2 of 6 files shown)


src/Factory/AppRouterFactory.php
```diff
@@ -7,14 +7,23 @@
...
 use App\Controller\SiteController;
+use Middlewares\BasicAuthentication;
+use Middlewares\DigestAuthentication;
+use App\Http\MiddlewareChain;
 
 class AppRouterFactory
 {
     public function __invoke(ContainerInterface $container)
     {
+        $basicAuth = $container->get(BasicAuthentication::class);
```
samdark: What if I need two different username/passwords for two different URLs?

razonyang: According to [BasicAuthentication::login](https://github.com/middlewares/http-authentication/blob/a91a5909ec6be91192dbfaccc487afcd434be58a/src/BasicAuthentication.php#L48), it seems impossible to do that, unless extends and overwrite it, or we can implement our own middleware.

src/Http/MiddlewareChain.php
```diff
@@ -0,0 +1,79 @@
...
+     */
+    public function __construct(MiddlewareInterface ... $middlewares)
+    {
+        $this->middlewares = $middlewares;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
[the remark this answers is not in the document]
razonyang: Yes, it was inspired by `MiddlewareDispatcher`.

src/Factory/AppRouterFactory.php
samdark: I don't think it is impossible. Likely, using yiisoft/factory or simply using `new` instead of getting `BasicAuthentication` from container for these would allow setting up multiple instances of `BasicAuthentication` with different `users` constructor parameter.

samdark:
Another common use-case is adding basic auth for a group of URLs. For example, securing an API i.e. everything that is starting from `/api` should be secured. @razonyang do you see a simple way to achieve that?

src/Http/MiddlewareChain.php
```diff
@@ -0,0 +1,79 @@
...
+
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * MiddlewareChain is a chaining of middlewares that implements MiddlewareInterface
+ */
+class MiddlewareChain implements MiddlewareInterface
```
samdark: This one should go into yiisoft/yii-web.

samdark: I'll handle adding it.

razonyang:
> Another common use-case is adding basic auth for a group of URLs. For example, securing an API i.e. everything that is starting from `/api` should be secured. @razonyang do you see a simple way to achieve that?

I have no idea, it is a responsibility of (group)router, but I don't know how to achieve it so far.

samdark:
Alright. I think the same, didn't want to voice it before hearing _maybe different_ opinion.

src/Http/MiddlewareChain.php
samdark: https://github.com/yiisoft/yii-web/commit/727a39171c395616cf694092403a01e608d1ed24

samdark:
Looks like it works quite well. `Chain` was merged and I don't think it makes sense to merge the rest. I should turn "using middlewares" into docs though.

samdark:
Thank you very much! I'll handle the docs.

samdark:
https://github.com/yiisoft/docs/blob/master/guide/en/structure/middleware.md

PR #5, opened 2019-07-08
fix error "Return value must be of the type string, null returned"

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |


samdark:
Merged. Thanks!

PR #6, opened 2019-07-11
Remove yiisoft/yii-base-web

damasco:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  |

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,6 @@
         "yiisoft/view": "^3.0@dev",
         "yiisoft/router": "^3.0@dev",
         "yiisoft/router-fastroute": "^3.0@dev",
-        "yiisoft/yii-base-web": "^3.0@dev",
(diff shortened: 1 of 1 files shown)


samdark:
Technically could be done but I need to re-check it. Likely it's a better idea to move some parts of the demo to common application base.

damasco:
Both of them are projects, I think you should not include one in the other.

samdark:
@hiqsol I think we need you :)

thenotsoft:
I think everything is better in one package, since only one layout is planned, or is the basic and adavance layout planned?

samdark:
No, there is no plan to have two application templates. The @hiqsol idea was that application template should be update-able i.e. you inherit from a template with your application.

samdark:
Closing for now.

PR #14, opened 2019-07-21
Add mailer example

razonyang:


src/Controller/SiteController.php
```diff
@@ -33,6 +52,55 @@ public function index(): ResponseInterface
...
     }
 
+    public function contact(RequestInterface $request): ResponseInterface
+    {
+        $parameters = [];
+        if ($request->getMethod() === 'POST') {
+            $config = array_merge(
+                require $this->aliases->get('@root/config/params.php'),
+                require $this->aliases->get('@root/config/params-local.php'),
+            );
```
machour: Is there a way to get these from composer-config-plugin instead?

params-local.php is meant to be optional during `composer du` but here it would break.. using the composer-config-plugin compiled output seems like a better idea, but I'm not really familiar with its API.

src/Controller/SiteController.php
```diff
@@ -33,6 +52,55 @@ public function index(): ResponseInterface
...
+            $sent = false;
+            $error = '';
+            try {
+                foreach (['subject', 'name', 'email', 'content'] as $name) {
+                    if (empty($_POST[$name])) {
```
machour: Shouldn't we be using PSR7 request object features instead?

views/site/contact.php
```diff
@@ -0,0 +1,36 @@
...
+<div class="alert alert-danger" role="alert">
+  <?= $error ?>
+</div>
+<?php endif; ?>
+<?php endif; ?>
+
+<form id="contactForm" method="POST" action="/contact" enctype="multipart/form-data">
+  <div class="form-group">
+    <label for="subject">Subject</label>
+    <input type="text" class="form-control" name="subject" id="subject" placeholder="Subject" value="<?= $_POST['subject'] ?? '' ?>" required>
```
machour: XSS ?

src/Controller/SiteController.php
```diff
@@ -1,26 +1,45 @@
 <?php
 namespace App\Controller;
 
+use Psr\Log\LoggerInterface;
```
machour: `psr/log` should be added to composer.json `require` section

src/Controller/SiteController.php
```diff
@@ -33,6 +52,55 @@ public function index(): ResponseInterface
         return $response;
     }
 
+    public function contact(RequestInterface $request): ResponseInterface
+    {
+        $parameters = [];
+        if ($request->getMethod() === 'POST') {
```
samdark: There is handy `Method::POST` constant for the purpose. Could be used.

src/Controller/SiteController.php
samdark: Yes:

```php
$config = require Builder::path('params');
```

It should load full config from `vendor/hiqdev/composer-config-plugin-output/params.php`.

src/Controller/SiteController.php
samdark: Yes, we should.

views/site/contact.php
```diff
@@ -0,0 +1,36 @@
+<?php if (isset($sent)): ?>
+<?php if($sent): ?>
+<div class="alert alert-success" role="alert">
+  Thanks to contact us, we'll get in touch with you as soon as possible.
+</div>
+<?php else: ?>
+<div class="alert alert-danger" role="alert">
+  <?= $error ?>
```
samdark: Should be encoded.

views/site/contact.php
samdark: Yes.

samdark:
@razonyang looks great overall but still needs some polishing. If you prefer, I can do it myself.

Also, seems we're lacking HTML helper that's useful for views. Going to extract it soon.

razonyang:
@samdark I've fixed some minor problems, as to the rest of problems(lack of enough knowledge and good idea), I need your help with them, you can do it yourself, thank you :smile:

razonyang:
Btw, I met some problems during the time of writing this example, it seems hard to retrieve some services, such as `logger`, `mailer` since `yii\helpers\Yii` was removed.

For example, I prefer to separate `form` logic from controller like this:

```php
public function contact()
{
    $form = new ContactForm();
    $response = $form->handle();
    // ...
}

class ContactForm
{
    // What is the proper way to inject dependencies here?
}
```

Another missing feature of this example is that it doesn't perform any validations(such as `email`). :(

samdark:
Logger and Mailer are supposed to be injected via constructors. Either you need to register ContactForm in DI container (then dependencies will be injected automatically) or pass instances from controller:

```php
public function contact(Mailer $mailer, Logger $logger)
{
    $form = new ContactForm($mailer, $logger);
    $response = $form->handle();
    // ...
}

class ContactForm
{
    public function __construct(Mailer $mailer, Logger $logger)
    {
          // ...
    }
}
```

samdark:
Merged! Moved contact form into its own controller afterwards.

samdark:
Thank you very much.

PR #16, opened 2019-08-02
Fix 'dd' function conflicts with dependencies

KartaviK:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️/❌

The meaning of the function name "dd" is "dump and die". Also in many cases needs dump more than one variable


samdark:
Merged. Thanks!

PR #23, opened 2019-09-08
Using Cycle ORM

samdark:
In this pull request we're using [Cycle ORM](https://github.com/cycle/orm) to store user data.

There's a console command `./vendor/bin/yii user/create` to create a user. After user is created you can try logging in with credentials entered.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -


FabrizioCaldarelli:
Why to use Cycle ORM instead Yii orm ?

lubosdz:
Same reason as usually - reduce maintenance costs due to insufficient funding of core developers :-(

demonking:
Have we a comparison between yii-orm and Cycle-orm?  

Usabilty,Performance etc?

lubosdz:
@demonking
No, not againts Yii2 ORM, only comparison Cycle vs Doctrine (symfony) & Eloquent (laravel):
https://github.com/cycle/docs/issues/3

FabrizioCaldarelli:
> Same reason as usually - reduce maintenance costs due to insufficient funding of core developers :-(

By proceeding in this way, Yii will become a set of external libraries and we will dependent from all of these. If they change something, we mandatory need to change. Bad thing.

terabytesoftw:
With yii3 you can use any library including Active Record of Yii2, the point is that you have to migrate it and that will take time, you can contribute helping in the migration.

demonking:
But it's not that easy with Yii ORM and Cycle ORM, the usage is different.

You cannot switch them without problems.

Cycle uses Annotation, Yii AR don't (or i haven't seen it).

samdark:
Seems there are questions about our plans:

1. @terabytesoftw is correct that we aim to make Yii work with any database abstraction layer be it Cycle, Doctrine, plain PDO or anything else.
2. @lubosdz is correct that maintenance cost of Yii 2 AR was quite high when it was developed actively. We were able to handle it with active participation of @cebe, @SilverFire, @klimov-paul, @dynasource, @arogachev. And, of course, with @qiangxue outstanding effort. So, indeed, we need more active developers to have our own database abstraction layer and it can be achieved in many ways including more funding.
3. We have no published comparison of Cycle / Yii 2 AR but I helped forming Cycle feature-set and query builder syntax at early stages so it's good enough. You can check the guide.
4. @FabrizioCaldarelli is correct that it's a good idea to have our own implementation if/when we can do better than the rest of implementations. If not, we introduce interfaces or, in rare cases, use third party libraries directly.
5. Each ORM syntax is different and, indeed, it's best to choose one at the start of the project and stick to it. But some patterns could help. In this pull request you can see repository in action. Note that no code working with IdentityRepository was changed.

wolfy-j:
Is it very easy to make Cycle build without Annotations (they are specifically moved out of core). We would def consider alternative ways of mapping definition if someone can propose Yii-like alternative.

rustamwin:
> So, indeed, we need more active developers to have our own database abstraction layer and it can be achieved in many ways including more funding.

@samdark :hand:

lubosdz:
@samdark 
You are in a difficult position. Would it not be safer to start with Yii3 only once sufficient funding is secured? Until then just maintaining Yii2 & develop solid business model ... OpenColletive may never provide enough funds to cover 4-5 developers. It's been a few months already but only 1 fulltime developer collected so far.

terabytesoftw:
I sincerely believe that this discussion has left the subject, simply Yii 3 will not be equal to Yii 2, and understanding what samdark says, everything will be migrated, everything will also depend on the help of the community.

samdark:
Yes, let's not off-topic here. If you have ideas to discuss, let's use forums.

roxblnfk:
>  Change default from plural to singular in table names

This behavior is hard-coded [here](https://github.com/cycle/annotated/blob/master/src/Entities.php#L174). 

If needed singular in table names then only manually set this in the annotations

wolfy-j:
@roxblnfk check 2.0.2, you can now configure the naming strategy for tables. https://github.com/cycle/annotated/blob/master/src/Entities.php#L31

roxblnfk:
@wolfy-j wow, my comment is outdated in just 18 minutes :)

samdark:
All code migrated to yiisoft/yii-cycle#1

PR #25, opened 2019-10-15
Cycle orm migrations

roxblnfk:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | no test
| Fixed issues  | #23 

Added new commands:
```bash
  migrate/generate  Generates a migration
  migrate/up        Execute all new migrations
  migrate/down      Rollback last migration
  migrate/list      Print list of all migrations
```


config/common.php
```diff
@@ -49,6 +52,23 @@
...
     MailerInterface::class => new MailerFactory(),
 
+
+    // Cycle DBAL
+    DatabaseManager::class => new CycleDbalFactory(),
     // Cycle ORM
     ORMInterface::class => new CycleOrmFactory(),
+
+    // Entity dir
+    # todo: replace to class object
```
samdark: Do you mean a factory?

roxblnfk: I'm not sure. Maybe here needed make a repository. Like EntityFileRepository that will contain paths to all Entities

roxblnfk:
Now i'm satisfied. Please let's review this

src/Console/Migration/.gitignore
```diff
@@ -0,0 +1 @@
+!.gitignore
```
samdark: How about moving migrations out of `src/Console`? These could be treated as non-source i.e. be in the `migrations` at the top level or at least could be moved out of `Console`.

samdark:
@roxblnfk I've pushed some minor adjustments. Overall looks good for me except where migrations are located.

src/Console/Migration/.gitignore
roxblnfk: Maybe you're right. In addition, I also wanted to make a drop cache when applying / rolling back migrations

roxblnfk: About the location of the migrations:
Migrations change (add) with Entities. I think they have the right to be in Src along with Entities

samdark: OK but, I think, not in Console.

roxblnfk: `src/Migration` ?

src/Console/Command/MigrateGenerate.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function configure(): void
+    {
+        $this->setDescription('Generates a migration');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->cycleHelper->generateMigration($this->migrator, $this->config);
```
samdark: @wolfy-j suggested that [such check is needed](https://github.com/spiral/framework/blob/master/src/Command/Cycle/MigrateCommand.php#L62).

wolfy-j: Cycle will generate migrations based on diff with current DB schema if the schema is not consistent you will generate broken migrations. This is a small nasty issue, so such check is recommended.

roxblnfk: I understand and support this recommendation. Also, it would be nice not to generate empty migrations

samdark: Hmm... Why would empty migrations be generated?

wolfy-j: We are checking generated migrations via Generator observer: https://github.com/spiral/framework/blob/master/src/Command/Cycle/MigrateCommand.php#L78

You are free to copy its code, it also gives you a nice description of what have changed.

roxblnfk: @samdark for example: if Entities not changed then running `migrate/generate` it creates migration contains void methods (`up()` and `down()` )
@wolfy-j ok, I'll see it later, and if necessary, i'll implement the comparison code

wolfy-j: Fixed in 1.0.5

wolfy-j:
I would recommend a way to add empty migration manually. There are always cases where you can't rely on ORM to solve diff.

src/Console/Command/MigrateGenerate.php
samdark: One question. How would it work if you write migration manually? Would that still create dummy migration classes for me

wolfy-j: ORM and manual migrations works using the same engine, you can combine them if you want. Basically ORM simply writes migrations for you, the rest is the same.

wolfy-j: https://github.com/cycle/migrations/blob/master/src/GenerateMigrations.php#L87

roxblnfk:
> I would recommend a way to add empty migration manually. There are always cases where you can't rely on ORM to solve diff.

Then you will have to change the code again, because now using \Cycle\Migrations\GenerateMigrations you can not create empty migrations

wolfy-j:
You can not create empty relations for ORM, you still can create migrations directly. Migrator and ORM/Migrations are two different packages.

That's how you can register migration: 
https://github.com/cycle/migrations/blob/master/src/GenerateMigrations.php#L87

And that's how you can generate empty one: 
https://github.com/cycle/migrations/blob/master/src/GenerateMigrations.php#L117

We have a command which allows us to scaffold migrations with some tables and columns pre-defined.

roxblnfk:
@wolfy-j  I just want to avoid code duplication

wolfy-j:
Well, you can always use string body like a classic template.

roxblnfk:
> I'm just saying that empty/user migration is nothing more than just a class in migrations directory.

Yes, that’s obvious. That is why it is useful to check for unapplied migrations before generating new ones. Because if you do not apply the previously generated migration, the new migration will take the all unapplied changes

roxblnfk:
All code migrated to https://github.com/yiisoft/yii-cycle/pull/1

PR #47, opened 2020-01-26
Clean and format

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/config/common.php
+++ b/config/common.php
@@ -19,7 +19,6 @@ use Yiisoft\Mailer\MailerInterface;
 return [
     Aliases::class => $params['aliases'],
     Psr\SimpleCache\CacheInterface::class => ArrayCache::class,
-    CacheInterface::class => Cache::class,
     Parameters::class => static function () use ($params) {
         return new Parameters($params);
     },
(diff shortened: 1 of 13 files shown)


config/common.php
```diff
@@ -19,7 +19,6 @@
 return [
     Aliases::class => $params['aliases'],
     Psr\SimpleCache\CacheInterface::class => ArrayCache::class,
-    CacheInterface::class => Cache::class,
```
samdark: Why was it removed?

xepozz: Duplicated https://github.com/yiisoft/yii-demo/blob/master/config/common.php#L26 https://github.com/yiisoft/yii-demo/blob/master/config/common.php#L22

samdark:
Merged.

config/common.php
roxblnfk: Duplicate came from this commit https://github.com/yiisoft/yii-demo/commit/ed718934c090a15f61686aa7893b5463e8516082
And i phink we had to remove
```php
    CacheInterface::class => [
        '__class' => Cache::class,
        '__construct()' => [
            '__class' => ArrayCache::class,
        ],
    ],
```
Because `\Yiisoft\Cache\Cache` requires `\Psr\SimpleCache\CacheInterface` in the constructor
Definition for `\Psr\SimpleCache\CacheInterface` declared in previous line:
```php
    // with namespaces
    \Psr\SimpleCache\CacheInterface::class => \Yiisoft\Cache\ArrayCache::class,
    \Yiisoft\Cache\CacheInterface::class => \Yiisoft\Cache\Cache::class,
```

PR #51, opened 2020-02-07
Simplify parameters

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ?
| Fixed issues  | -

### DIFF
--- a/src/Parameters.php
+++ b/src/Parameters.php
@@ -4,9 +4,7 @@ declare(strict_types=1);
 
 namespace App;
 
-use function array_key_exists;
-use function strpos;
-use function explode;
+use Yiisoft\Arrays\ArrayHelper;
 
 /**
  * Parameters provides a way to get application parameters defined in config/params.php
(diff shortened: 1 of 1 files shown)


thenotsoft:
Your version works with both `'test'.key' => 'value'` and `'test' => ['key' => 'value']` ?

rustamwin:
> Your version works with both 'test'.key' => 'value' and 'test' => ['key' => 'value'] ?

Yes. It works as well

thenotsoft:
I'm not sure. When calling `->get('test.key')` in the first case it will return value, in seconda case will return null or default.

roxblnfk:
> I'm not sure. When calling `->get('test.key')` in the first case it will return value, in seconda case will return null or default.

https://github.com/yiisoft/arrays/blob/64c6972bdd6fb804e92f70aca30202630b719635/src/ArrayHelper.php#L169

thenotsoft:
> > I'm not sure. When calling `->get('test.key')` in the first case it will return value, in seconda case will return null or default.
> 
> https://github.com/yiisoft/arrays/blob/64c6972bdd6fb804e92f70aca30202630b719635/src/ArrayHelper.php#L136

Yes, you're right, I didn’t look at the sources of the method (

samdark:
Thanks!

Issue #52, opened 2020-02-09
Change behaviour of `render` method


xepozz:
### What steps will reproduce the problem?
Using `render` function to render content of view file looks too hard.
Instead of return the result of `render` function we should to create response before and write data after.
Let's delegate creating the `Response` object to `render` function. 

### What is the expected result?
User controllers will be more simpler and cleaner
```php
public function index(): Response
{
    return $this->render('index');
}
```

### What do you get instead?
```php
public function index(): Response
{
    $output = $this->render('index');
    $response = $this->responseFactory->createResponse();
    $response->getBody()->write($output);
    return $response;
}
```

samdark:
View component is not aware of request-response. Its job is rendering templates. I don't think it's a good idea to couple it like suggested.

Still, I agree that it looks a bit cumbersome to create response like that (while it makes perfect sense if redirect or sending headers is involved).

Possible solutions:

1. Allow returning strings from handlers (actions).
2. Provide a shortcut method like createResponseWithBody($string).
3. ???

xepozz:
If you want to render and collect result to string you can use `$this->getView()->render(...);`
`Controller::render()` is only wrapper over `$view->render`. I do not see any problems in returning  `Response`

Moreover, such a template is familiar to all PHP developers (at least), because all frameworks provide such a convenient interface.

samdark:
Returning response directly from `render()`? Some references would not hurt...

yiiliveext:
I think @xepozz mean something like this
```php
//abstract class Controller
protected function render(string $view, array $parameters = []): string
{
    $response = $this->responseFactory->createResponse();
    $content = $this->view->render($view, $parameters, $this);
    
    return $response->getBody()->write($this->renderContent($content));
}
```

samdark:
Yes but that would be:

```php
protected function render(string $view, array $parameters = []): Response
```

yiiliveext:
> Yes but that would be:
> 

Of course, I've just forgotten to change after copying.

samdark:
Thought more about it. It doesn't sound bad to do it.

Issue #53, opened 2020-02-09
Transfer base controller to the yii-web package


xepozz:
I am sure that many users do not need to configure the controller for themselves at all.

And those who need it will be able to inherit the basic controller and supplement it, or create their own.

Also since the class is abstract, please rename it `AbstractController`

samdark:
I think it's not the best idea. While base controller might be convenient, I think it's much better to have a set of traits in `yii-web` such as `ViewRendererConventionTrait` that you can either add to your handler (action, controller) directly or use in your own base controller (if you prefer inheritance).

xepozz:
This controller may be useful in another modules (e.g. `yii-debug`, `yii-gii`, `yii-apidoc` and etc.)

>  ViewRendererConventionTrait that you can either add to your handler

How will you inject dependencies in this case (e.g. `Renderer`, `Flash`, `UrlGenerator` and etc.)?

samdark:
Yes, good question.

Overall, idea of base controller is controversial: https://twitter.com/sam_dark/status/1227328185680977921

To me, it's convenient and I don't see any critical drawbacks despite it being formally not correct.

xepozz:
And what's the solution? 
We can create it, but user will decide whether to use it.

yiiliveext:
I think the abstract controller should be in the app template. There is no need to move it to the yii-web package.

roxblnfk:
I think that the controller sould be moved, but not now. The controller is not yet complete. When he ceases to be a draft and finds a stable state, then he will have to be transferred
I also hope that we will have several for different tasks.

samdark:
Yes, we can wait with this one.

Another possibility from that Twitter thread is to wrap view-resolving into its own class and inject it as a dependency.

armpogart:
I'm for having `AbstractController` in app templates, that will allow to have different (specialized) `AbstractController`s in each template with right dependencies for each case (e.g. ApiController doesn't need view injection), and this approach will allow for easier customization of Base (Abstract) controller as it will be directly visible to end users (developers). This will keep framework away from recommending best or better practices for application code directly and just a propose a way to do something in app template and any developer is free to use the template and or build his/her own.

But it is crucial to have an AbstractController in template at least, as some of my beginner developers (in studio I own) were not sure what to do with all this packages and how to render views after using yii2 extensively and AbstractController will just make their lives easier.

roxblnfk:
[...]

PR #54, opened 2020-02-13
Render now returns ResponseInterface instead of string

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | #52

### DIFF
--- a/config/common.php
+++ b/config/common.php
@@ -19,7 +19,6 @@ use Yiisoft\Mailer\MailerInterface;
 return [
     Aliases::class => $params['aliases'],
     Psr\SimpleCache\CacheInterface::class => ArrayCache::class,
-    CacheInterface::class => Cache::class,
     Parameters::class => static function () use ($params) {
         return new Parameters($params);
     },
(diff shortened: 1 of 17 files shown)


roxblnfk:
I think it is wrong that the render method returns a `ResponseInterface`.
I suggest considering another way.
Let the `ActionCaller` wrap the string  value in a `ResponseInterface`.

samdark:
> I think it is wrong that the render method returns a ResponseInterface.

I was thinking the same before taking into account that it's controller's render method, not a method of a View.

roxblnfk:
> > I think it is wrong that the render method returns a ResponseInterface.
> 
> I was thinking the same before taking into account that it's controller's render method, not a method of a View.

It is also worth considering that this controller does not implement `MiddlewareInterface`, therefore it is not required to return a `ResponseInterface`. This also applies to other `Actions` passed to `ActionCaller`.

samdark:
> Let the ActionCaller wrap the string value in a ResponseInterface.

What if you need headers as well? In case of `render()` returning `ResponseInterface` you can add headers after the rendering.

roxblnfk:
> What if you need headers as well? In case of `render()` returning `ResponseInterface` you can add headers after the rendering.

Just like now, when we can manually create a Response.

But what if the Action specified in the ActionCaller returns a string rather than ResponseInterface?
What if I want to treat the rendered page as a string? Should I unpack it from Response? (I remind you that ResponseFactory writes a line to a tmp file when creating a stream)
I would be glad to see a separate method like `wrapToResponse ()`, but not so that render transformed it as Response by default

yiiliveext:
> What if I want to treat the rendered page as a string? Should I unpack it from Response? (I remind you that ResponseFactory writes a line to a tmp file when creating a stream)

You still can do it, just render it manually
```php
$content = $this->view->render($view, $parameters, $this);
$bodyText = $this->renderContent($content);
```

samdark:
Let's have it. We're likely refactor base controllers in something else anyway: https://github.com/yiisoft/yii-demo/issues/53

@xepozz thanks!

PR #57, opened 2020-02-19
Moved composer-config-output directory to our own `runtime` folder

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

At now hiqdev/composer-config-plugin will write into our own `runtime` directory instead of `vendor/hiqdev/composer-config-plugin/composer-config-output`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,6 +62,7 @@
         "branch-alias": {
             "dev-master": "3.0.x-dev"
         },
+        "config-plugin-output-dir": "runtime/config-plugin-output-dir",
         "config-plugin": {
             "common": "config/common.php",
             "params": [
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "branch-alias": {
             "dev-master": "3.0.x-dev"
         },
-        "config-plugin-output-dir": "runtime/cache/config",
+        "config-plugin-output-dir": "runtime/build/config",
         "config-plugin": {
             "common": "config/common.php",
             "params": [
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "branch-alias": {
             "dev-master": "3.0.x-dev"
         },
-        "config-plugin-output-dir": "runtime/config-plugin-output-dir",
+        "config-plugin-output-dir": "runtime/cache/config",
         "config-plugin": {
             "common": "config/common.php",
             "params": [


schmunk42:
About about `config/_generated`?

xepozz:
> About about `config/_generated`?

Maybe `runtime/cache/config`?

rustamwin:
> Maybe runtime/cache/config?

maybe `runtime/generated/config`?

xepozz:
What the purpose of the `generated`? All things in `runtime` are autogenerated

rustamwin:
> What the purpose of the generated? All things in runtime are autogenerated

Ah.. right

schmunk42:
Things in `runtime` are writable by the webserver, the application configuration should not be modifiable by default.

In our setup I'd also generate the config during build and/or even commit it.

xepozz:
> Things in `runtime` are writable by the webserver, the application configuration should not be modifiable by default.
> 
> In our setup I'd also generate the config during build and/or even commit it.

We can improve the plugin so that it can set permissions

schmunk42:
I'd also not treat those files as typical runtime files, since you should usually be able to clean your runtime folder and stuff gets regenerated if required.

In general, how should that be handled in production? Commit files (or CI build with Docker), manual creation with composer after git update / FTP upload, auto-generation from `index.php`?

xepozz:
I suggest to have 2 ways to solve your task:
1. Console command `cache/warmup`.
    After deploy we run this command and all cache files are generating.
2. Runtime generating
    If original file was changed, cache must be generated at this moment.

samdark:
@xepozz re-generation of configs in runtime is expensive operation.

xepozz:
> @xepozz re-generation of configs in runtime is expensive operation.

It's better, than 500 error :)

composer.json
```diff
@@ -62,6 +62,7 @@
         "branch-alias": {
             "dev-master": "3.0.x-dev"
         },
+        "config-plugin-output-dir": "runtime/cache/config",
```
samdark: How about runtime/build instead of runtime/cache? That is to give a hint that a build is required in order to get the data back if it's deleted.

xepozz: `runtime/build/config`?

samdark: yes

xepozz: ok

samdark:
Merged. Thanks!

PR #59, opened 2020-02-20
Lazy Rendering

roxblnfk:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

I want to introduce experiments on lazy rendering.

## How it works now:

- We receive a HTTP request and translate it into a Request object
- We send the Request through middleware to the controller
- There may be some action on data change
- Now we need to prepare the contents of the page. **We designate the page and call `render()`**
- The page content is rendered and written to the buffer.
- Then a layout with markers is rendered into which the page content is written. The layout content is also written to the buffer
- on a calling `WebView::endPage()` the buffer is retrieved, the markers are replaced with CSS/JS.\
  The result is written as a string into stream of a Response object.
- Response is piped back
- Emitter receives a Response and emit content to the client

## The idea of lazy rendering:

- When we call the `render()` method, a generator is created. Rendering as such does not occur.
- The generator is wrapped in a stream and placed in a Response object.
- Response is piped back to Emitter
- Emitter starts emitting the contents of the stream to the client, and here inside the generator, rendering starts
- When ready, data is immediately sent to the client, not to the buffer

Visually the essence of lazy rendering in this picture:
image

## Features

1. Page does not linger in buffer
2. Stream of the Response object does not know the length of the content
3. At the time of rendering, all the middleware is already “behind”

### Benefits

- From feature №1 follows less memory consumption, and you can also generate a page of any length
- From item 1 follows a shorter waiting time to first byte (**TTFB**)
- Now we are using PSR-7 from `nyholm`, which means we can avoid unnecessary I/O (Response from `nyholm` writes the buffer to a tmp file that uses as stream)

### Disadvantages

- From feature №3, difficulties may arise with middleware who write information related to rendered content in HTTP headers \
  *Solution*: we can force the rendering to run at any stage by simply requesting content from the stream
- From feature №3, in the case of an uncaught error at the stage of rendering and lack of buffering, it will not be possible to change the HTTP code to 500. At the same time, part of the page can already be generated and sent.
- Current WebView is completely unsuitable for this approach. In addition, without markers that do not work without page prerendering, the Asset system may work incorrectly (some Assets may be declared while the page content is rendered, but we need to know about them earlier)

## Implementation

I put the experimental code in a separate folder `src/LazyRendering`
You can find the start links on the index page.
image

I also made some demonstration pages, and added different rendering modes for them:

- **Lazy Mode** - deferred rendering pages. Page renders as it emits.
- **Classic Mode** *(imitation)* - page is rendered and buffered before the response object is sent back via pipeline.
- **Combined Mode** - deferred rendering pages. First yielded value will be emitted immediately, the rest of the contents will be buffered and emitted in one piece.

I had to replace the existing WebView for these pages with a streaming counterpart.

## My results

Without I/O actions in a Controller
image

## Questions

- Do you see a benefit for yourself in this approach?
- What do you think about making streaming WebView?
- What do you think about lazy rendering?
- Would you like to see the minimum TTFB on your site?


samdark:
That's a good demo of current emitter capabilities. Potentially it could be converted into tutorial and/or parts of it could be made reusable. Need to check it in more detail.

xepozz:
I like your idea, but I don't understand the next disadvantages:
> - From feature №3, difficulties may arise with middleware who write information related to rendered content in HTTP headers
> - From feature №3, in the case of an uncaught error at the stage of rendering and lack of buffering, it will not be possible to change the page code to 500. At the same time, part of the page can already be generated and sent.

Could you explain it more please ?

rustamwin:
Looks interesting :slightly_smiling_face:

src/LazyRendering/Http/BaseController.php
```diff
@@ -0,0 +1,134 @@
...
+        $actionMethod = $this->getActionMethod($request);
+        if ($actionMethod === null || !method_exists($this, $actionMethod)) {
+            return null;
+        }
+        $data = (new Injector($this->container))->invoke([$this, $actionMethod]);
+
+        $response = $data instanceof Response ? $data : $this->prepareResponse($data);
+
+        // Force Buffering (classic mode)
+        if (($this->request->getQueryParams()['forceBuffering'] ?? 0) === '1') {
```
samdark: What is `1`?

src/LazyRendering/Http/BaseController.php
```diff
@@ -0,0 +1,134 @@
...
+            // Buffering from generator
+            $content = $response->getBody()->getContents();
+            $stream = $this->streamFactory->createStream($content);
+            return $response->withBody($stream);
+        }
+
+        if (($this->request->getQueryParams()['forceBuffering'] ?? 0) === '2') {
```
samdark: What is `2`?

src/LazyRendering/Http/BaseController.php
```diff
@@ -0,0 +1,134 @@
...
+        if (!$page instanceof Generator) {
+            $page = (static function (iterable $iterable) {
+                yield from $iterable;
+            })($page);
+        }
+        // Add layout rendering
+        if ($this->pageLayout !== null) {
+            if (is_string($this->pageLayout)) {
+                $this->pageLayout = $this->container->get($this->pageLayout);
+            } elseif (!is_object($this->pageLayout) || !method_exists($this->pageLayout, 'render')) {
```
samdark: Looks like an interface would be a good idea here...

src/LazyRendering/Http/StreamedController.php
```diff
@@ -0,0 +1,145 @@
...
+
+class StreamedController extends BaseController
+{
+    public const ROUTE_NAME = 'lazy-render';
+
+    protected $pageLayout = MainLayout::class;
+
+    public function pageIndex(): iterable
+    {
+        foreach (get_class_methods($this) as $method) {
```
samdark: That's way too magical to understand it. How about calling all these methods explicitly for demo purpose instead?

src/LazyRendering/View/MainLayout.php
```diff
@@ -0,0 +1,65 @@
...
+            ob_start();
+        }
+        ob_end_clean();
+    }
+
+    protected function renderPage(iterable $content): Generator
+    {
+        $this->assetManager->register([AppAsset::class]);
+
+        echo '<html><head>' . ($this->title !== null ? '<title>' . Html::encode($this->title) . '</title>' : '');
```
samdark: Is mix of `echo` and `yield` intended?

roxblnfk: Yes
Each 'yield' means the end of a chunk in chucked output. I use 'echo' to combine several small chunks that don't use i/o into one

samdark:
While it's definitely interesting and allowed to test Yii 3 for edge cases, I don't think it should be merged into demo app.

PR #61, opened 2020-02-23
Enable debugger

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

This PR will be worked after merging https://github.com/yiisoft/yii-debug/pull/34


public/index.php
```diff
@@ -6,13 +6,23 @@
 use Yiisoft\Yii\Web\ServerRequestFactory;
 
 require_once dirname(__DIR__) . '/vendor/autoload.php';
+require_once dirname(__DIR__) . '/src/globals.php';
```
samdark: No, that is not desired since this way $container won't be available in `globals.php`.

xepozz: It's not used there.
Maybe `dd()` can be moved to `yiisoft/var-dumper` and initialized there?

samdark: > It's not used there.

Part of the demo, one day, will become a template. It will be used in apps for certain.

> Maybe dd() can be moved to yiisoft/var-dumper and initialized there?

These shortcuts, while handy, aren't really a good practice. For now I'd avoid having these all over the place.

xepozz: > These shortcuts, while handy, aren't really a good practice. For now I'd avoid having these all over the place.

Then why `$container` should be available in `globals.php`?

samdark: https://github.com/yiisoft/di/issues/99

xepozz: Looks like a part of `Yii2` design... 
And `Laravel`'s facades...

samdark: Yes, kind of. That is hacky but handy.

samdark:
Merged. 👍

PR #65, opened 2020-02-28
Separate emitter from the Application

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | -

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -2,7 +2,9 @@
 
 use hiqdev\composer\config\Builder;
 use Yiisoft\Di\Container;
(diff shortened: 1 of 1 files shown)


roxblnfk:
Do we need `EmitterInterface` in this case?

samdark:
@roxblnfk good question. Maybe we do not...

terabytesoftw:
I like the interface if I want to use another no emitter package I could do it

PR #66, opened 2020-03-04
Fix access to runtime

Fantom409:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

When run application via docker-compose.yml Builder throw Exception due to insufficient access to runtime folder. This PR fix that

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
     },
     "scripts": {
         "post-install-cmd": [
-            "chmod -R 0777 runtime"
+            "php -r \"chmod('runtime',0777);\""
          ]
     },
     "extra": {
--- a/composer.json
+++ b/composer.json
@@ -59,6 +59,11 @@
             "App\\": "src"
         }
     },
+    "scripts": {
+        "post-install-cmd": [
+            "chmod -R 0777 runtime"
+         ]
+    },
     "extra": {
         "branch-alias": {
             "dev-master": "3.0.x-dev"
--- a/composer.json
+++ b/composer.json
@@ -61,8 +61,8 @@
     },
     "scripts": {
         "post-install-cmd": [
-            "php -r \"chmod('runtime',0777);\""
-         ]
+            "App\\Installer::postInstall"
+        ]
     },
     "extra": {
         "branch-alias": {
--- a/composer.json
+++ b/composer.json
@@ -60,8 +60,8 @@
         }
     },
     "scripts": {
-        "post-install-cmd": [
-            "App\\Installer::postInstall"
+        "post-update-cmd": [
+            "App\\Installer::postUpdate"
         ]
     },
     "extra": {
--- a/src/Installer.php
+++ b/src/Installer.php
@@ -20,10 +20,7 @@ final class Installer
     {
         chmod($path, $mode);
         $iterator = new RIterator(
-            new DirIterator(
-                $path,
-                FSIterator::SKIP_DOTS | FSIterator::CURRENT_AS_PATHNAME
-            ),
+            new DirIterator($path, FSIterator::SKIP_DOTS | FSIterator::CURRENT_AS_PATHNAME),
             RIterator::SELF_FIRST
         );
         foreach ($iterator as $item) {
--- a/src/Installer.php
+++ b/src/Installer.php
@@ -11,7 +11,8 @@ class Installer
         self::rchmod('runtime', 0777);
     }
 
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -59,6 +59,11 @@
             "App\\": "src"
         }
     },
+    "scripts": {
+        "post-install-cmd": [
+            "chmod -R 0777 runtime"
```
samdark: That would not work for installing the package in Windows.

Fantom409:
Last commit doesn't solve problem - because chmod should apply recursively

src/Installer.php
```diff
@@ -0,0 +1,24 @@
...
+use Composer\Script\Event;
+
+class Installer
+{
+    public static function postInstall(Event $event): void
+    {
+        self::rchmod('runtime', 0777);
+    }
+
+    private static function rchmod(string $path, int $mode): void
```
roxblnfk:
```suggestion
    private static function chmodRecursive(string $path, int $mode): void
```

src/Installer.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace App;
+
+use Composer\Script\Event;
+
+class Installer
+{
+    public static function postInstall(Event $event): void
```
roxblnfk:
```suggestion
    public static function postInstall(Event $event = null): void
```
Because we don't use this event

src/Installer.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace App;
+
+use Composer\Script\Event;
+
+class Installer
```
roxblnfk:
```suggestion
final class Installer
```

src/Installer.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
```
roxblnfk:
```suggestion

declare(strict_types=1);

```

src/Installer.php
```diff
@@ -0,0 +1,28 @@
...
+    public static function postUpdate(Event $event = null): void
+    {
+        self::chmodRecursive('runtime', 0777);
+    }
+
+    private static function chmodRecursive(string $path, int $mode): void
+    {\
+        $dir = new \DirectoryIterator($path);
+        foreach ($dir as $item) {
+            if (!$item->isDot()) {
```
roxblnfk:
```suggestion
            if ($item->isDot()) {
                continue;
            }
```

roxblnfk:
what about using [RecursiveDirectoryIterator](https://www.php.net/RecursiveDirectoryIterator)?

src/Installer.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static function postUpdate(Event $event = null): void
+    {
+        self::chmodRecursive('runtime', 0777);
+    }
+
+    private static function chmodRecursive(string $path, int $mode): void
+    {
+        chmod($path, $mode);
+        $iterator = new RecursiveIteratorIterator(new \RecursiveDirectoryIterator($path), RecursiveIteratorIterator::SELF_FIRST);
```
roxblnfk:
```suggestion
        $iterator = new RecursiveIteratorIterator(
            new \RecursiveDirectoryIterator(
                $path,
                \FilesystemIterator::SKIP_DOTS | \FilesystemIterator::CURRENT_AS_PATHNAME
            ),
            RecursiveIteratorIterator::SELF_FIRST
        );
        foreach ($iterator as $item) {
            chmod($item, $mode);
        }
```

roxblnfk: you can also add short aliases for longnamed classes like `use FilesystemIterator as FSIterator; use RecursiveDirectoryIterator as DirIterator`...

src/Installer.php
```diff
@@ -0,0 +1,33 @@
...
+    public static function postUpdate(Event $event = null): void
+    {
+        self::chmodRecursive('runtime', 0777);
+    }
+
+    private static function chmodRecursive(string $path, int $mode): void
+    {
+        chmod($path, $mode);
+        $iterator = new RIterator(
+            new DirIterator(
```
roxblnfk:
```suggestion
            new DirIterator($path, FSIterator::SKIP_DOTS | FSIterator::CURRENT_AS_PATHNAME),
```
Could you do it in one line?
With short names it does not longer than 120 characters

roxblnfk:
Merged. Thanks!

PR #72, opened 2020-03-26
Deferred response demo

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

PR #73, opened 2020-03-27
Deferred response with response formatter

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Alternative to https://github.com/yiisoft/yii-demo/pull/72


src/Controller.php
```diff
@@ -33,11 +34,12 @@ public function __construct(
...
-        $response = $this->responseFactory->createResponse();
-        $content = $this->view->render($view, $parameters, $this);
-        $response->getBody()->write($this->renderContent($content));
+        $controller = $this;
+        $contentRenderer = function () use ($view, $parameters, $controller) {
+            return $controller->renderContent($controller->view->render($view, $parameters, $controller));
+        };
 
-        return $response;
+        return $this->factory->create(DeferredResponse::class, [$contentRenderer]);
```
samdark: Likely could be made more explicit about dependencies:

1. Create `DefferedResponseFactory` that creates `DeferredResponse`.
2. Inject it into constructor.

src/Factory/AppRouterFactory.php
```diff
@@ -38,12 +42,21 @@ public function __invoke(ContainerInterface $container)
...
                 Route::get('/{login}', [UserController::class, 'profile'])
                      ->name('user/profile'),
             ]),
 
+            // User
+            Group::create('/api', [
+                Route::get('/user', [ApiUserController::class, 'index'])
+                    ->name('api/user/index'),
+                Route::get('/user/{login}', [ApiUserController::class, 'profile'])
+                    ->addMiddleware(new ResponseFormatter($container->get(JsonResponseFormatter::class)))
```
samdark: Do we need non-deferred responses at all?

src/Controller.php
yiiliveext: Done

.gitignore
```diff
@@ -25,3 +25,5 @@ phpunit.phar
 /vendor
 /composer.lock
 /node_modules
+
+package.json
```
rustamwin: Why?
```suggestion
package-lock.json
```

yiiliveext: Scrutinizer generates an error with it. Temporary disabled. This file is regenerated every time on the composer update.

src/DeferredResponseFactory.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public function createResponse(int $code = 200, string $reasonPhrase = '', $data = null): ResponseInterface
+    {
+        return new DeferredResponse($data, $this->createResponseInternal($code, $reasonPhrase), $this->streamFactory);
+    }
+
+    private function createResponseInternal(int $code = 200, string $reasonPhrase = ''): ResponseInterface
+    {
+        if (2 > \func_num_args()) {
```
samdark: How does that work? There are exactly two arguments for `createResponseInternal()` passed so the check should never be true.

src/HtmlResponseFormatter.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * @var string the XML encoding. If not set, it will use the value of [[Response::charset]].
+     */
+    private string $encoding = 'UTF-8';
+
+    public function format(DeferredResponse $deferredResponse): ResponseInterface
+    {
+        $data = $deferredResponse->getData();
+        $response = $deferredResponse->getResponse();
+        $response->getBody()->write($data);
```
samdark: Should we cast `$data` to string here?

src/DeferredResponseFactory.php
yiiliveext: This a copy of the code from Psr17Factory

samdark: I don't understand it.

src/HtmlResponseFormatter.php
yiiliveext: Done

samdark:
Do not merge. See https://github.com/yiisoft/yii-web/pull/233

PR #74, opened 2020-03-27
add missing install step composer install

philippfrenzel:
otherwise no vendor directory will contain the server

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌


samdark:
Good catch. Thanks.

PR #80, opened 2020-04-02
Update for EventManager

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Merge after https://github.com/yiisoft/event-dispatcher/pull/19


public/index.php
```diff
@@ -1,36 +1,35 @@
...
+use Yiisoft\Composer\Config\Builder;
 use Yiisoft\Di\Container;
 use Yiisoft\Http\Method;
 use Yiisoft\Yii\Web\Application;
 use Yiisoft\Yii\Web\SapiEmitter;
 use Yiisoft\Yii\Web\ServerRequestFactory;
 
 require_once dirname(__DIR__) . '/vendor/autoload.php';
-
+require_once dirname(__DIR__) . '/src/globals.php';
```
samdark, line 12: It should not be moved since container won't be available in `globals.php` this way.

public/index.php
```diff
@@ -1,36 +1,35 @@
...
 // Don't do it in production, assembling takes it's time
 Builder::rebuild();
-
+$startTime = microtime(true);
```
samdark, line 15: https://github.com/yiisoft/yii-demo/blob/master/src/Timer.php

PR #84, opened 2020-04-22
reconstruct App\Controller->render make it easy to understand.

dvaknheo:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | 

`$controller` => `$self` .  just for replace keyword `this` to `self` for closure . reduce to think `$controller` may has other useful thing.

```
return $controller->renderContent($controller->view->render($view, $parameters, $controller));
 ```
=>
```
return $self->renderProxy($view, $parameters); 
```

// closure for proxy,use parent var. 
// reduce a level(hierarchy?) .
```
if ($layout === null) {
    return $content;
}
```
// return early . reduce a thinking stack.

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -46,7 +46,7 @@ abstract class Controller implements ViewContextInterface
         $content = $this->view->render($view, $parameters, $this);
         $user = $this->user->getIdentity();
         $layout = $this->findLayoutFile($this->layout);
-        
+
         if ($layout === null) {
             return $content;
         }
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -33,10 +33,7 @@ abstract class Controller implements ViewContextInterface
 
     protected function render(string $view, array $parameters = []): ResponseInterface
     {
-        $self = $this;
-        $contentRenderer = static function () use ($view, $parameters, $self) {
-            return $self->renderProxy($view, $parameters);
-        };
+        $contentRenderer = fn () => $this->renderProxy($view, $parameters);
 
         return $this->responseFactory->createResponse($contentRenderer);
     }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dvaknheo** [comment]
em .  styleci check failed , because a blank line .
this is my first use pull request to github .
how to i do next ? fixed and pushed .
mey delete and then create annother pr ? or do any thing other?

File: src/Controller.php [hunk 1]
```diff
@@ -33,31 +33,31 @@ public function __construct(
...
+        $self = $this;
+        $contentRenderer = static function () use ($view, $parameters, $self) {
+            return $self->renderProxy($view, $parameters);
         };
 
         return $this->responseFactory->createResponse($contentRenderer);
     }
 
-    private function renderContent($content): string
+    private function renderProxy($view, $parameters): string
```
xepozz, line 44: Parameters should be typed

src/Controller.php
```diff
@@ -33,31 +33,31 @@ public function __construct(
 
     protected function render(string $view, array $parameters = []): ResponseInterface
     {
-        $controller = $this;
-        $contentRenderer = static function () use ($view, $parameters, $controller) {
-            return $controller->renderContent($controller->view->render($view, $parameters, $controller));
+        $self = $this;
+        $contentRenderer = static function () use ($view, $parameters, $self) {
+            return $self->renderProxy($view, $parameters);
         };
```
xepozz, line 39:
```suggestion
        $contentRenderer = fn () => $this->renderProxy($view, $parameters);
```

src/Controller.php
dvaknheo: ok, done.

samdark:
Thanks!

PR #85, opened 2020-04-22
Change cache type

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/config/common.php
+++ b/config/common.php
@@ -5,6 +5,12 @@ use App\Factory\MailerFactory;
 use App\Parameters;
 use App\Timer;
 use Psr\Log\LoggerInterface;
+use Yiisoft\Aliases\Aliases;
+use Yiisoft\Cache\File\FileCache;
+use Psr\Container\ContainerInterface;
+use Psr\SimpleCache\CacheInterface;
+use Yiisoft\Cache\Cache;
+use Yiisoft\Cache\CacheInterface as YiiCacheInterface;
 use Yiisoft\Log\Target\File\FileRotator;
 use Yiisoft\Log\Target\File\FileRotatorInterface;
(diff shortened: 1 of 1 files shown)


config/common.php
```diff
@@ -17,8 +23,10 @@
 $timer->start('overall');
 
 return [
-    \Psr\SimpleCache\CacheInterface::class => \Yiisoft\Cache\ArrayCache::class,
-    \Yiisoft\Cache\CacheInterface::class => \Yiisoft\Cache\Cache::class,
+    CacheInterface::class => function (ContainerInterface $container) {
+        return new FileCache($container->get(Aliases::class)->get('@runtime/cache'));
+    },
+    YiiCacheInterface::class => Cache::class,
```
roxblnfk, line 29: I phink full class name better than alias in this case

config/common.php
```diff
@@ -17,8 +23,10 @@
 $timer->start('overall');
 
 return [
-    \Psr\SimpleCache\CacheInterface::class => \Yiisoft\Cache\ArrayCache::class,
-    \Yiisoft\Cache\CacheInterface::class => \Yiisoft\Cache\Cache::class,
+    CacheInterface::class => function (ContainerInterface $container) {
```
roxblnfk, line 26:
```suggestion
    CacheInterface::class => static function (ContainerInterface $container) {
```

samdark:
👍

config/common.php
samdark: c565c6b71f2c5a2797846b52f992dc69bd346d6f

PR #86, opened 2020-04-23
Fix repositories for pgsql

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/Blog/Archive/ArchiveRepository.php
+++ b/src/Blog/Archive/ArchiveRepository.php
@@ -95,7 +95,7 @@ final class ArchiveRepository
             $str = ['year' => '%Y', 'month' => '%m', 'day' => '%d'][$attr];
             return new Fragment("strftime('{$str}', post.published_at) {$attr}");
         }
-        return new Fragment("extract({$attr} from post.published_at) {$attr}");
+        return new Fragment("extract({$attr} from post.published_at) \"{$attr}\"");
     }
 
     private function getDriver(): DriverInterface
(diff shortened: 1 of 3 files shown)


src/Blog/Tag/TagRepository.php
```diff
@@ -69,7 +69,7 @@ public function getTagMentions(int $limit = 0): DataReaderInterface
             ->columns(['t.label', 'count(*) count'])
             ->innerJoin('post', 'p')->on('p.id', 'postTag.post_id')->onWhere(['p.public' => true])
             ->innerJoin('tag', 't')->on('t.id', 'postTag.tag_id')
-            ->groupBy('tag_id');
+            ->groupBy('t.label, tag_id');
```
xepozz, line 72: Can you also append a table name to `tag_id`?

src/Blog/Archive/ArchiveRepository.php
```diff
@@ -95,7 +95,7 @@ private function extractFromDateColumn(string $attr): Fragment
             $str = ['year' => '%Y', 'month' => '%m', 'day' => '%d'][$attr];
             return new Fragment("strftime('{$str}', post.published_at) {$attr}");
         }
-        return new Fragment("extract({$attr} from post.published_at) {$attr}");
+        return new Fragment("extract({$attr} from post.published_at) \"{$attr}\"");
```
xepozz, line 98: How about to quote the second usage $attr?

yiiliveext: The first {$attr} has a fixed set of values and is internal for SQL.

PR #89, opened 2020-04-24
Add .env db params

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️


yiiliveext:
> Can we use another filename than `.env` eg. `app.env`, because `docker-compose` also reads those files by default and uses them in it's control environment, **not** in the container unless you expose them.
> 
> This kinda sucks, but it won't be changed, I guess, since there were already huge discussions about it.

See https://github.com/yiisoft/composer-config-plugin/pull/44

samdark:
I don't think it makes sense to merge this one before https://github.com/yiisoft/composer-config-plugin/issues/47 is solved.

samdark:
For now, I'm closing this one...

Issue #50, opened 2020-01-29
To abstract src/Parameters.php to separate package


xepozz:
I think https://github.com/yiisoft/yii-demo/blob/master/src/Parameters.php should be moved outside from the `yiisoft/yii-demo` package.
It can be a separate packet, like as the `yiisoft/arrays` or the `yiisoft/arrays` or can be added to already existsing packages, but I don't sure.
Also, would be nice to make an interface for that service.

roxblnfk:
I think using `Parameters` as a whole is bad practice and its use must be minimized

xepozz:
Why not? With using `Parameters` we can do different environments for the project: dev, prod, stage, test, etc.
Using `Parameters` we can take enviroment variable (i.e. from .env file).

thenotsoft:
> I think using `Parameters` as a whole is bad practice and its use must be minimized

What do you think is better to pass parameters?

> Using Parameters we can take enviroment variable (i.e. from .env file).

I think for this it is better to create a separate class `Environment`

xepozz:
> I think for this it is better to create a separate class Environment

Yes of course.
I suggest to make an interface for this classes. 
For example, I can use only `.env` files to configure my application, and you can use only `config/params.php` for this, both via `ParametersInterface` (I think it's worth working on a name :))

thenotsoft:
Yeah, I use "Params" and "Environment" and it is very convenient to use.
It would be nice to move it in a separate package, and add functionality if necessary, and will already be used at the discretion of the developer.

yiiliveext:
I think an interface for parameters would be the better choice.

samdark:
We have composer-config-plugin that already solves environment problem.

xepozz:
At now we have at least 2 usages: in [`yii-gii`](https://github.com/yiisoft/yii-gii/blob/d949d2fdfdb3615c03633c037001c1af09f2c78f/src/Parameters.php) and [`yii-web`](https://github.com/yiisoft/yii-demo/blob/master/src/Parameters.php)
Further there will be more. We just must create new component for it.

samdark:
Decided not to do it. See #88.

PR #91, opened 2020-04-28
Fix #88, #50: Eliminate Parameters service

### LINKED ISSUE #88: Do not access parameters directly
Parameters should be used to configure services and should not be accessed directly.

https://github.com/yiisoft/yii-demo/blob/master/src/Parameters.php should be eliminated.

samdark:
- Eliminate Parameters service
- Move contact to its own namespace
- Introduce debug mode for mailer

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #88, #50


src/Contact/ContactController.php
```diff
@@ -57,34 +55,19 @@ public function contact(ServerRequestInterface $request): ResponseInterface
...
-                    );
+                    $message->addFile($files['file']);
                 }
 
-                $message->send();
+                $this->mailer->send($message);
+
                 $sent = true;
             } catch (\Throwable $e) {
+                throw $e;
```
gegorov2030, line 70: По ошибке попало?

samdark: Yes.

PR #96, opened 2020-05-07
Add route/list command

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | -
| Fixed issues  | -


src/Command/Route/ListCommand.php
```diff
@@ -0,0 +1,66 @@
...
+            ->setDescription('List all registered routes')
+            ->setHelp('This command displays a list of registered routes.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $table = new Table($output);
+        $io = new SymfonyStyle($input, $output);
+        /** @var Route[] $routes */
+        $routes = $this->urlMatcher->getRouteCollection()->getRoutes();
```
yiiliveext, line 41: Maybe it will be better to use RouteCollection()->getRouteTree(false) to get routes tree

xepozz: @yiiliveext or we need to drop `getRouteTree` method from `RouteCollection` instead?

src/Command/Route/ListCommand.php
```diff
@@ -0,0 +1,66 @@
...
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Yiisoft\Router\Route;
+use Yiisoft\Router\UrlMatcherInterface;
+use Yiisoft\Yii\Console\ExitCode;
+
+class ListCommand extends Command
+{
+    private UrlMatcherInterface $urlMatcher;
+
+    protected static $defaultName = 'route/list';
```
xepozz, line 21: I think names `router/list` or `routes/list` woulbe be more relevant

src/Command/Route/ListCommand.php
```diff
@@ -0,0 +1,66 @@
...
+        $this
+            ->setDescription('List all registered routes')
+            ->setHelp('This command displays a list of registered routes.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $table = new Table($output);
+        $io = new SymfonyStyle($input, $output);
+        /** @var Route[] $routes */
```
xepozz, line 40: It's redundant, because `getRoutes` must be typed as `Route[]`.
See https://github.com/yiisoft/router/pull/48

src/Command/Route/ListCommand.php
```diff
@@ -0,0 +1,66 @@
...
+            );
+            if (next($routes) !== false) {
+                $table->addRow(new TableSeparator());
+            }
+        }
+
+        try {
+            $table->render();
+        } catch (\Throwable $exception) {
+            $io->error($exception->getMessage());
```
xepozz, line 61: Please log full error trace. 
```suggestion
            $io->error((string) $exception);
```

src/Command/Route/ListCommand.php
```diff
@@ -0,0 +1,66 @@
...
+            return $exception->getCode() ?: ExitCode::UNSPECIFIED_ERROR;
```
xepozz, line 62: `$exception->getCode()` isn't suitable to system codes of processes endings

src/Command/Route/ListCommand.php
```diff
@@ -0,0 +1,66 @@
...
+                    $route->getPattern(),
+                    implode(',', $route->getDefaults())
+                ]
+            );
+            if (next($routes) !== false) {
+                $table->addRow(new TableSeparator());
+            }
+        }
+
+        try {
```
xepozz, line 58: Why do you do it?
The rendering table of routes is target goal of this command. 
It should be done or crashed with error and full stacktrace

src/Command/Route/ListCommand.php
yiiliveext: > @yiiliveext or we need to drop `getRouteTree` method from `RouteCollection` instead?

Routes tree is more informative and convenient.

src/Command/Route/ListCommand.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $table = new Table($output);
+        $io = new SymfonyStyle($input, $output);
+        /** @var Route[] $routes */
+        $routes = $this->urlMatcher->getRouteCollection()->getRoutes();
+        $table->setHeaders(['Host', 'Methods', 'Name', 'Pattern', 'Defaults']);
+        foreach ($routes as $key => $route) {
```
xepozz, line 43: Also please sort routes by host/name/method

xepozz: `$key` is never used

src/Command/Route/ListCommand.php
rustamwin: > Maybe it will be better to use RouteCollection()->getRouteTree(false) to get routes tree

This is a route table, not a tree.

src/Command/Route/ListCommand.php
rustamwin: Hmm.. makes sense.

src/Command/Route/ListCommand.php
rustamwin: Probably the name `routes / list` might be appropriate, but we accept the singular instead of the plural.

xepozz: How about `router/list`? In the future it can be multiple commands like the following: `router/match`, `router/group`, `router/middlewares`, `router/handler` and etc.

rustamwin: Done.

samdark:
Good one. Thanks!

PR #101, opened 2020-05-28
getToken() never return null

andrew-demb:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔


samdark:
That is correct. Thanks!

Issue #102, opened 2020-05-31
yii serve gives error if debug is disabled


siripravi:
### What steps will reproduce the problem?

1. Remove `'debugger.enabled' => true,` from params.php
2. When I run the command 'yii serve', I get the following error

### What do you get instead?

![yiierror](https://user-images.githubusercontent.com/174201/83344657-7deda600-a327-11ea-84ce-7b6c613b6c76.PNG)

### Additional info

| Q                | A
| ---------------- | ---
| Version          | 1.0.?
| PHP version      | 
| Operating system |

siripravi:
![image](https://user-images.githubusercontent.com/174201/83344676-c6a55f00-a327-11ea-9828-9938231e83f9.png)

siripravi:
PHP Fatal error:  Uncaught Yiisoft\Factory\Exceptions\NotFoundException: No defi
nition for Yiisoft\Yii\Debug\Storage\StorageInterface in D:\Yii3Projects\yii-dem
o\vendor\yiisoft\di\src\Container.php:233
Stack trace:
#0 D:\Yii3Projects\yii-demo\vendor\yiisoft\di\src\Container.php(209): Yiisoft\Di
\Container->buildPrimitive('Yiisoft\\Yii\\Deb...')
#1 D:\Yii3Projects\yii-demo\vendor\yiisoft\di\src\Container.php(155): Yiisoft\Di
\Container->buildInternal('Yiisoft\\Yii\\Deb...')
#2 D:\Yii3Projects\yii-demo\vendor\yiisoft\di\src\Container.php(88): Yiisoft\Di\
Container->build('Yiisoft\\Yii\\Deb...')
#3 D:\Yii3Projects\yii-demo\vendor\yiisoft\factory\src\Definitions\ClassDefiniti
on.php(36): Yiisoft\Di\Container->get('Yiisoft\\Yii\\Deb...')
#4 D:\Yii3Projects\yii-demo\vendor\yiisoft\factory\src\Definitions\ArrayBuilder.
php(109): Yiisoft\Factory\Definitions\ClassDefinition->resolve(Object(Yiisoft\Di
\Container))
#5 D:\Yii3Projects\yii-demo\vendor\yiisoft\factory\src\Definitions\ArrayBuilder.
php(93): Yiisoft\Factory\Definitions\ArrayBuilder- in D:\Yii3Projects\yii-demo\v
endor\yiisoft\di\src\Container.php on line 233

[...]

PR #103, opened 2020-06-01
fix: yii serve gives error #102

forecho:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #102

### DIFF
--- a/config/params.php
+++ b/config/params.php
@@ -4,6 +4,7 @@ use App\Command;
 use Cycle\Schema\Generator;
 
 return [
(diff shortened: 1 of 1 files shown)


samdark:
Doesn't seem to fix it.

samdark:
Thanks. We have to find out what's wrong with debug disabled though...

PR #108, opened 2020-06-24
Improve controller name definition

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️


src/Contact/ContactController.php
```diff
@@ -30,11 +30,6 @@ public function __construct(
         parent::__construct($responseFactory, $user, $aliases, $view);
     }
 
-    protected function getId(): string
-    {
-        return 'contact';
```
samdark, line 35: Wouldn't new one result in `contact/contact`?

src/Blog/BlogController.php
```diff
@@ -21,11 +21,6 @@ final class BlogController extends Controller
     private const POPULAR_TAGS_COUNT = 10;
     private const ARCHIVE_MONTHS_COUNT = 12;
 
-    protected function getId(): string
-    {
-        return 'blog';
```
samdark, line 26: Wouldn't it result it `blog\blog`?

src/Controller.php
```diff
@@ -66,7 +68,7 @@ private function renderProxy(string $view, array $parameters = []): string
 
     public function getViewPath(): string
     {
-        return $this->aliases->get('@views') . '/' . $this->getId();
+        return $this->aliases->get('@views') . '/' . self::getName();
```
samdark, line 71:
```suggestion
        return $this->aliases->get('@views') . '/' . static::getName();
```

xepozz:
I need to copy-paste changes from https://github.com/yiisoft/app/pull/21

src/Blog/BlogController.php
xepozz: No
image

src/Contact/ContactController.php
xepozz: No
image

samdark:
:+1:

PR #135, opened 2020-09-08
Creating tags, adding/removing tags from posts

Fantom409:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/views/blog/post/__form.php
+++ b/views/blog/post/__form.php
@@ -44,10 +44,10 @@ if (!empty($error ?? '')) {
 
     <div class="form-group">
         <label for="AddTag">Add tag</label>
-        <input type="text" class="form-control" id="AddTag" placeholder="Add tag" value="">
+        <input type="text" class="form-control" id="addTag" placeholder="Add tag" value="">
         <?= Html::button(
     'Add',
-    ['class' => 'btn btn-primary mt-2', 'onclick' => 'addTag()']
+    ['class' => 'btn btn-primary mt-2', 'id' => 'addTagButton']
 ) ?>
     </div>
 
--- a/views/blog/post/__form.php
+++ b/views/blog/post/__form.php
@@ -6,6 +6,8 @@
  * @var $body array
  * @var $csrf string
  * @var $action string
+ * @var $tags array
+ * @var $title string
  */
 
 use Yiisoft\Html\Html;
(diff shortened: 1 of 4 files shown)


views/blog/post/__form.php
```diff
@@ -41,5 +48,61 @@
...
+                }());
+            JS
+            ]
+) ?>
+    </div>
+
+    <div class="form-group" id="tags">
+        <?php foreach ($tags as $tag): ?>
+             <span class="badge badge-info mr-2" id="tag<?= $tag ?>">
+                <input type="hidden" name="tags[]" value="<?= $tag ?>">
```
samdark, line 89: All values outputted in HTML context should be HTML-encoded else you'll get XSS.

```suggestion
                <input type="hidden" name="tags[]" value="<?= Html::encode($tag) ?>">
```

views/blog/post/__form.php
```diff
@@ -41,5 +48,61 @@
                   placeholder="Content" required><?= Html::encode($body['content'] ?? '') ?></textarea>
     </div>
 
+    <div class="form-group">
+        <label for="AddTag">Add tag</label>
+        <input type="text" class="form-control" id="AddTag" placeholder="Add tag" value="">
+        <?= Html::button(
+    'Add',
+    ['class' => 'btn btn-primary mt-2', 'onclick' => <<<JS
```
samdark, line 56: It is better to have it in a separate JS-file. Likely, asset bundles is a good match to register it: https://github.com/yiisoft/app/tree/master/src/Asset

samdark (changes requested):

views/blog/post/__form.php
```diff
@@ -41,5 +42,35 @@
                   placeholder="Content" required><?= Html::encode($body['content'] ?? '') ?></textarea>
     </div>
 
+    <div class="form-group">
+        <label for="AddTag">Add tag</label>
+        <input type="text" class="form-control" id="AddTag" placeholder="Add tag" value="">
+        <?= Html::button(
+    'Add',
+    ['class' => 'btn btn-primary mt-2', 'onclick' => 'addTag()']
```
samdark, line 50: It is better not to use `onclick`. Instead, try addEventListener in `app.js` along with adding ID to the button:

```javascript
document.getElementById("#addTag").element.addEventListener('click', addTagFunction);
```

samdark (changes requested):

samdark (approved):

samdark:
Thanks!

PR #138, opened 2020-09-13
Refactoring post controller

romkatsu:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | Refactoring post controller


src/Blog/Post/PostService.php
```diff
@@ -0,0 +1,48 @@
...
+use App\Blog\Entity\Tag;
+use App\Blog\Tag\TagRepository;
+use App\Entity\User;
+
+final class PostService
+{
+    private PostRepository $repository;
+    private TagRepository $tagRepository;
+
+    public function __construct(PostRepository $repository, TagRepository $tagRepository)
```
yiiliveext, line 17: Maybe it will be more convenient.
```php
public function __construct(PostRepository $repository, TagRepository $tagRepository, UserService $userService)
{
    $this->repository = $repository;
    $this->tagRepository = $tagRepository;
    $this->userService = $userService;
}

public function savePost(Post $model, PostForm $form, User $user = null): void
{
    if ($user === null) {
        if ( $this->userService->getUser() === null)
        {
             throw new \InvalidArgumentException('User not set');
        }
        
        $user = $this->userService->getUser();
    }
    //....
}
```

romkatsu: This method should not be called if there is no user.

yiiliveext: > This method should not be called if there is no user.

This is the way to set the current user as default.

romkatsu: > > This method should not be called if there is no user.
> 
> This is the way to set the current user as default.

I don't think it's necessary.

src/Factory/AppRouterFactory.php
```diff
@@ -76,10 +79,12 @@ public function __invoke(ContainerInterface $container)
...
                 // Add Post page
                 Route::methods([Method::GET, Method::POST], '/page/add', [PostController::class, 'add'])
-                    ->name('blog/add'),
+                    ->name('blog/add')->addMiddleware(Authentication::class),
                 // Edit Post page
                 Route::methods([Method::GET, Method::POST], '/page/edit/{slug}', [PostController::class, 'edit'])
-                    ->name('blog/edit'),
+                    ->name('blog/edit')
+                    ->addMiddleware(Authentication::class)
+                    ->addMiddleware(fn () => $container->get(AccessChecker::class)->withPermission('editPost')),
```
yiiliveext, line 87:
```sugestion
->addMiddleware(fn (AccessChecker $checker) => $checker->withPermission('editPost')),
```

romkatsu: fixed

samdark (approved):

samdark:
👍

PR #170, opened 2020-11-10
Fix config console.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/config/providers-web.php
+++ b/config/providers-web.php
@@ -3,14 +3,7 @@
 declare(strict_types=1);
 
 use Yiisoft\Arrays\Modifier\ReverseBlockMerge;
-use Yiisoft\Composer\Config\Builder;
-use Yiisoft\Yii\Event\EventDispatcherProvider;
 
 return [
-    'yiisoft/event-dispatcher/eventdispatcher' => [
-        '__class' => EventDispatcherProvider::class,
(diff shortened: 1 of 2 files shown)


terabytesoftw:
Before merge: https://github.com/yiisoft/yii-event/pull/9

config/providers-web.php
```diff
@@ -3,14 +3,7 @@
...
-use Yiisoft\Composer\Config\Builder;
-use Yiisoft\Yii\Event\EventDispatcherProvider;
 
 return [
-    'yiisoft/event-dispatcher/eventdispatcher' => [
-        '__class' => EventDispatcherProvider::class,
-        '__construct()' => [Builder::require('events-web')]
-    ],
-
     ReverseBlockMerge::class => new ReverseBlockMerge(),
```
viktorprogger, line 8: Is it necessary?

viktorprogger (changes requested):

viktorprogger:
There are some more config arrays with an only `ReverseBlockMerge` element. Can we remove that element in this PR?

config/providers-web.php
samdark: Yes.

samdark (approved):

PR #184, opened 2020-12-01
Move C3.php to index-test.php

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -28,7 +28,7 @@ build:
                 override:
                     -
                         command: |
-                            php -S 127.0.0.1:8080 -t public > runtime/yii.log 2>&1 &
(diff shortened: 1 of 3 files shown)


.scrutinizer.yml
```diff
@@ -28,7 +28,7 @@ build:
                 override:
                     -
                         command: |
-                            php -S 127.0.0.1:8080 -t public > runtime/yii.log 2>&1 &
+                            php -S 127.0.0.1:8080 -t public public/index-test.php > ./runtime/yii.log 2>&1 &
```
samdark, line 31: It can be done with `./yii serve 127.0.0.1:8080 --router=public/index-test.php > ./runtime/yii.log 2>&1 &`. That should give equal result but may be better for education purposes.

samdark (approved):

PR #186, opened 2020-12-02
Group by usecase

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -


config/params.php
```diff
@@ -102,8 +102,7 @@
             ],
         ],
         'annotated-entity-paths' => [
-            '@src/Entity',
-            '@src/Blog/Entity',
+            '@src',
```
BoShurik, line 105: Is there a performance impact? Cycle needs to check all classes in that case

samdark: Sure there is but that's done only once. Then schema is cached.

BoShurik: Is it possible to warmup cache?

src/User/Console/CreateCommand.php
```diff
@@ -2,9 +2,9 @@
 
 declare(strict_types=1);
 
-namespace App\Command\User;
+namespace App\User\Console;
```
BoShurik, line 5: Why commands inside a `Console` namespace, but controllers does not have own `Controller`?
Maybe it worse to add additional level of context: `Console`, `Controller`, `Entity`, `Repository` etc

samdark: Yes, makes sense when there are multiple objects of the same type.

BoShurik: Same for https://github.com/yiisoft/yii-demo/tree/group-by-usecase/src/Blog

config/params.php
samdark: Yes, with `./vendor/bin/yii cycle/schema/rebuild`.

rustamwin:
[Group classes by use-case · Issue #3 · yiisoft/app-api](https://trello.com/c/cWSKLufg/156-group-classes-by-use-case-issue-3-yiisoft-app-api)

Issue #200, opened 2020-12-15
Argument 1 passed to Yiisoft\Html\Html::span() must be of the type string, int given


1Luc1:
### What steps will reproduce the problem?
- get latest commit
- click "Blog" link

### What is the expected result?
showing the blog page

### What do you get instead?
Error message:
```
TypeError: Argument 1 passed to Yiisoft\Html\Html::span() must be of the type string, int given, called in /srv/http/yii3/views/blog/_topTags.php on line 35 and defined in /srv/http/yii3/vendor/yiisoft/html/src/Html.php:1406
Stack trace:
#0 /srv/http/yii3/views/blog/_topTags.php(35): Yiisoft\Html\Html::span()
#1 /srv/http/yii3/vendor/yiisoft/view/src/PhpTemplateRenderer.php(13): require('/srv/http/yii3/...')
...
```

### Additional info
| Q                | A
| ---------------- | ---
| Version          | master (fb5464b)
| PHP version      |  PHP: v7.4.13
| Operating system | Manjaro 20.2

### Possible Solution ?
within `_topTags.php`
` $count = (string) $tagValue['count'];`

and `_archive.php`
`$count = (string) $item['count'];`

samdark:
@Mister-42 would you please take a look?

Mister-42:
I did not encounter this error, but indeed int is not valid for Html::span and similar tags. Casting to string works in such cases.

Mister-42:
@1Luc1 Can you do a PR for those 2 lines?

1Luc1:
> I did not encounter this error, but indeed int is not valid for Html::span and similar tags. Casting to string works in such cases.
Interesting, does this have to do something with the php version?

> @1Luc1 Can you do a PR for those 2 lines?
Yes I can.

Mister-42:
> Interesting, does this have to do something with the php version?
Not between 7.4 and 8 as far as I know.

> Yes I can.
Thank you. I'd do it myself if I wasn't at work, so you helping out speeds it up 👍🏻

Mister-42:
@samdark Issue can be closed.

PR #201, opened 2020-12-15
fixed bug where param for Html::span was int

1Luc1:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #200


samdark:
👍

PR #202, opened 2020-12-16
Add visual indicator to labels of required fields

Mister-42:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  |


Mister-42:
Test on contact page: this change works, but there are empty labels on the page. These are now made visible by the presence of a red star.
`<label class="control-label control-label required" for="contactform-subject"></label>`

Investigating

src/Contact/ContactForm.php
```diff
@@ -19,7 +19,7 @@ final class ContactForm extends FormModel
     public function attributeLabels(): array
     {
         return [
-            'username' => 'Username',
+            'username' => 'Name',
```
samdark, line 22:
```suggestion
            'name' => 'Name',
```

views/layout/main.php
```diff
@@ -42,6 +42,12 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Yii Demo</title>
+    <style>
+        label.required:not(:empty):after {
```
samdark, line 46: It's better to place it into `site.css`.

samdark (changes requested):

samdark:
👍

Issue #192, opened 2020-12-07
Add 404 fallback page


Mister-42:
### What steps will reproduce the problem?
Visit a page that does not exist

### What is the expected result?
A nice 404 page. Bonus points for cats 😁 

### What do you get instead?
We were unable to find the page /x.

### Additional info

| Q                | A
| ---------------- | ---
| Version          | dev
| PHP version      | 7.4.13
| Operating system | Debian 10

1Luc1:
I'm interested in how to achieve this and played around a little bit with it. I came up with creating an own `NotFoundHandler` which implements the `RequestHandlerInterface`. This new Handler will be passed as `fallbackHandler` within the Application creation. May this the preferred way?

samdark:
It was meant to be so. Is that convenient?

samdark:
https://github.com/yiisoft/app/blob/master/config/web/application.php#L28

1Luc1:
I see, i see. I can make a PR?

Mister-42:
> I see, i see. I can make a PR?

Yes, please 👍🏻

PR #203, opened 2020-12-16
Added 404 fallback page

1Luc1:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #192 

no cats, sry :smile_cat:

### DIFF
--- a/src/Handler/NotFoundHandler.php
+++ b/src/Handler/NotFoundHandler.php
@@ -21,8 +21,6 @@ class NotFoundHandler implements RequestHandlerInterface
 
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
-        return $this->viewRenderer
-            ->render('404', ['requestedUri' => (string) $request->getUri()])
-            ->withStatus(Status::NOT_FOUND);
+        return $this->viewRenderer->render('404')->withStatus(Status::NOT_FOUND);
     }
 }
--- a/src/Handler/NotFoundHandler.php
+++ b/src/Handler/NotFoundHandler.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Handler;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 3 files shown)


views/site/404.php
```diff
@@ -0,0 +1,31 @@
...
+
+use Yiisoft\Html\Html;
+?>
+
+<div class="card shadow p-5 my-5 mx-5 bg-white rounded">
+    <div class="card-body text-center ">
+        <h1 class="card-title display-1 fw-bold">404</h1>
+        <p class="card-text">
+            <?php echo "The page "
+                . Html::span($requestedUri, ['class' => 'text-muted'])
```
Mister-42, line 18:
```php
. Html::span($urlMatcher->getCurrentUri()->getPath(), ['class' => 'text-muted'])`
```

src/Handler/NotFoundHandler.php
```diff
@@ -0,0 +1,28 @@
...
+
+    public function __construct(ViewRenderer $viewRenderer)
+    {
+        $this->viewRenderer = $viewRenderer->withControllerName('site');
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        return $this->viewRenderer
+            ->render('404', ['requestedUri' => (string) $request->getUri()])
```
Mister-42, line 25: $urlMatcher is available in view, so no need to add parameter to view.

views/site/404.php
samdark: $requestedUri should be passed through `Html::encode()`.

Mister-42: > $requestedUri should be passed through `Html::encode()`.

I changed my proposal accordingly 👍🏻

Mister-42:
Hard way to learn `$urlMatcher->getCurrentRoute()` will be `NULL`. Makes sense, but something to be aware of when used in layout.

samdark (approved):

samdark:
Thanks.

PR #204, opened 2020-12-16
Activate menu items on every subpage as well

Mister-42:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | 

This code works, but is ugly. Anyone got something better?

### DIFF
--- a/views/layout/main.php
+++ b/views/layout/main.php
@@ -55,9 +55,9 @@ echo Nav::widget()
         ->options(['class' => 'navbar-nav mx-auto'])
         ->items(
             [
-                ['label' => 'Blog', 'url' => $urlGenerator->generate('blog/index')],
+                ['label' => 'Blog', 'url' => $urlGenerator->generate('blog/index'), 'active' => $urlMatcher->getCurrentRoute() !== null && dirname($urlMatcher->getCurrentRoute()->getName()) === 'blog'],
                 ['label' => 'Comments Feed', 'url' => $urlGenerator->generate('blog/comment/index')],
-                ['label' => 'Users', 'url' => $urlGenerator->generate('user/index')],
+                ['label' => 'Users', 'url' => $urlGenerator->generate('user/index'), 'active' => $urlMatcher->getCurrentRoute() !== null && dirname($urlMatcher->getCurrentRoute()->getName()) === 'user'],
                 ['label' => 'Contact', 'url' => $urlGenerator->generate('site/contact')],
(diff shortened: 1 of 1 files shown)


views/layout/main.php
```diff
@@ -55,9 +57,9 @@
         ->options(['class' => 'navbar-nav mx-auto'])
         ->items(
             [
-                ['label' => 'Blog', 'url' => $urlGenerator->generate('blog/index')],
+                ['label' => 'Blog', 'url' => $urlGenerator->generate('blog/index'), 'active' => in_array($currentRoute, ['blog/index', 'blog/add', 'blog/archive/index', 'blog/archive/month', 'blog/archive/year', 'blog/edit', 'blog/post', 'blog/tag'])],
```
samdark, line 60:
```suggestion
                ['label' => 'Blog', 'url' => $urlGenerator->generate('blog/index'), 'active' => StringHelper::startsWith($currentRoute, 'blog/')],
```

views/layout/main.php
```diff
@@ -55,9 +57,9 @@
...
                 ['label' => 'Comments Feed', 'url' => $urlGenerator->generate('blog/comment/index')],
-                ['label' => 'Users', 'url' => $urlGenerator->generate('user/index')],
+                ['label' => 'Users', 'url' => $urlGenerator->generate('user/index'), 'active' => in_array($currentRoute, ['user/index', 'user/profile'])],
```
samdark, line 62:
```suggestion
                ['label' => 'Users', 'url' => $urlGenerator->generate('user/index'), 'active' => StringHelper::startsWith($currentRoute, 'user/')],
```

views/layout/main.php
samdark: But don't forget to `use` the helper.

samdark (changes requested):
Using string match would allow you not to modify condition when you add another sub-route.

views/layout/main.php
Mister-42: Actually, that would also pull in `blog/comment/index`, which I excluded on purpose (it's the next menu entry)

samdark: Well, for that `!== 'blog/comment/index'` condition could be added. Still better than keeping the list up to date.

Mister-42: Fair enough. I was not aware that part exists 👍🏻

Mister-42: Fair enough

samdark (approved):

samdark:
👍

PR #221, opened 2021-01-12
Clear entity mappers

roxblnfk:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

Cleared entity mappers from unnecessary methods that caused parasitic SQL queries (UPDATE after INSERT for each record)

image
image


samdark (approved):

samdark:
👍

PR #234, opened 2021-02-05
Remove unused yiisoft/var-dumper package

devanych:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,6 @@
         "yiisoft/serializer": "^3.0@dev",
         "yiisoft/strings": "^1.0",
         "yiisoft/user": "^3.0@dev",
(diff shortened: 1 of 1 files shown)


roxblnfk:
What about this?
https://github.com/yiisoft/yii-demo/blob/2a2414fb5ac67d7058724160baeaf9e6656335d1/src/functions.php#L1-L39

devanych:
> What about this?
> 
> https://github.com/yiisoft/yii-demo/blob/2a2414fb5ac67d7058724160baeaf9e6656335d1/src/functions.php#L1-L39

Woops! :)

PR #239, opened 2021-02-15
build: settings for initial launch

githubjeka:
My first init Yii3 app send me on page with An internal server error occurred and nothing else

image

Not very friendly :)


samdark (approved):

samdark:
👍

PR #240, opened 2021-02-15
fix: Uncaught TypeError

githubjeka:
Argument must be of type array.
`\Psr\Http\Message\ServerRequestInterface::getParsedBody` could be `null|object|array`


samdark (approved):

samdark:
Thanks.

PR #241, opened 2021-02-15
docs: Non-canonical order of @var

githubjeka:
Many tools expect correct syntax

https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#517-var
https://www.jetbrains.com/help/phpstorm/php-non-canonical-order-of-elements.html
https://phpstan.org/writing-php-code/phpdocs-basics

### DIFF
--- a/resources/mail/layouts/html.php
+++ b/resources/mail/layouts/html.php
@@ -1,5 +1,10 @@
 <?php
-/* @var $content string Mail contents as view render result */
+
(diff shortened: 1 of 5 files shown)


samdark (approved):

samdark:
Thanks.

PR #246, opened 2021-02-23
Add dependency checker

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -


samdark:
Makes it hard to define meta-packages such as yii-view and not require yiisoft/view. Won't adopt it.

PR #247, opened 2021-02-24
Add ApplicationRunner and delete src/functions.php

devanych:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -


xepozz (approved):

BoShurik:
Should cli entry point use this runner?

devanych:
> Should cli entry point use this runner?

If for tests, then yes. If we are talking about a console application, then no. The console application must use its own entry point.

samdark:
👍

PR #256, opened 2021-03-04
Added SQLite dependency to composer.json

Nex-Otaku:
SQLite extension is required to run demo app, so the extension must be set as dependency.

| Q             | A
| ------------- | ---
| Is bugfix?    | Yes
| New feature?  | No
| Breaks BC?    | No
| Fixed issues  | -

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -26,7 +26,6 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php }}"
-          extensions: sqlite3
           tools: composer:v2, cs2pr
           coverage: none
 
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -26,6 +26,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php }}"
+          extensions: sqlite3
           tools: composer:v2, cs2pr
           coverage: none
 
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -18,6 +18,7 @@
     "prefer-stable": true,
     "require": {
         "php": "^7.4|^8.0",
+        "ext-sqlite3": "^7.4|^8.0",
```
roxblnfk, line 21:
```suggestion
        "ext-sqlite3": "*",
```

roxblnfk (commented):
It can be more useful

composer.json
roxblnfk: Perhaps the pdo + pdo-sqlite should be required instead this

composer.json
```diff
@@ -18,6 +18,7 @@
     "prefer-stable": true,
     "require": {
         "php": "^7.4|^8.0",
+        "ext-sqlite3": "*",
```
roxblnfk, line 21:
```suggestion
        "ext-pdo_sqlite": "*",
```

.github/workflows/static.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php }}"
+          extensions: sqlite3
```
roxblnfk, line 29:
```suggestion
```

Nex-Otaku:
Checks failed.. (

PR #260, opened 2021-03-05
Remove dist config files

Mister-42:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |


Mister-42:
Not advisable when composer will report everything as changed :(

PR #289, opened 2021-05-02
fix: Call to undefined method Yiisoft\Profiler\Target\FileTarget::dis…

githubjeka:
image


roxblnfk (approved):

roxblnfk:
Fix is correct but merged in https://github.com/yiisoft/yii-demo/pull/291

PR #299, opened 2021-05-20
Remove default merge plan from configuration

devanych:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

Now no default merge plan is required, it will be generated when installing the app.


vjik:
This is not template, this is application... Merge plan should be in GIT of application.

devanych:
> This is not template, this is application... Merge plan should be in GIT of application.

What is it in GIT for?

vjik (approved):
While yii 3 not released it is easier

PR #304, opened 2021-05-27
Remove dependencie innecesary composer.json.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


vjik (approved):

vjik:
Dependency checker work only with composer 2, but demo application should work also on composer 1.

PR #327, opened 2021-08-04
New providers

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌


src/ApplicationRunner.php
```diff
@@ -22,6 +22,8 @@
 use Yiisoft\Yii\Web\Application;
 use Yiisoft\Yii\Web\SapiEmitter;
 use Yiisoft\Yii\Web\ServerRequestFactory;
+use Yiisoft\Yii\Debug\Proxy\ContainerProxy;
```
samdark, line 25: Looks like these aren't used.

samdark (approved):

PR #331, opened 2021-08-17
Adjust to router changes

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️/❌
| Fixed issues  | -

### DIFF
--- a/src/ViewInjection/ContentViewInjection.php
+++ b/src/ViewInjection/ContentViewInjection.php
@@ -6,21 +6,17 @@ namespace App\ViewInjection;
 
 use Yiisoft\Form\Widget\Field;
 use Yiisoft\Router\UrlGeneratorInterface;
-use Yiisoft\Router\UrlMatcherInterface;
 use Yiisoft\Yii\View\ContentParametersInjectionInterface;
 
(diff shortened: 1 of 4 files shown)


src/ViewInjection/ContentViewInjection.php
```diff
@@ -6,21 +6,17 @@
...
 use Yiisoft\Yii\View\ContentParametersInjectionInterface;
 
 class ContentViewInjection implements ContentParametersInjectionInterface
 {
-    private UrlMatcherInterface $urlMatcher;
     private UrlGeneratorInterface $url;
     private Field $field;
 
     public function __construct(
-        UrlMatcherInterface $urlMatcher,
```
samdark, line 19: Shouldn't we inject `CurrentRoute` here instead?

rustamwin: No. Already injected in `commonParameters` https://github.com/yiisoft/demo/pull/331/files#diff-b59c295812dde5ba7714e8519aac18c6ec217ddb1313c1bdbc7ba5c466d5261bR16

samdark (approved):

Issue #326, opened 2021-08-02
Add example for rate-limiter


xepozz:
https://github.com/yiisoft/rate-limiter

PR #337, opened 2021-08-23
feat: added rate limit

githubjeka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #326 

http://localhost/api/user

image

http://localhost/api/user

> Too Many Requests


src/User/Controller/RateLimitToUserApiMiddleware.php
```diff
@@ -0,0 +1,50 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Yii\RateLimiter\Counter;
+use Yiisoft\Yii\RateLimiter\LimitRequestsMiddleware;
+use Yiisoft\Yii\RateLimiter\Policy\LimitPerIp;
+use Yiisoft\Yii\RateLimiter\Storage\StorageInterface;
+
+class RateLimitToUserApiMiddleware implements MiddlewareInterface
```
rustamwin, line 17: What is the benefit if it already exists?

githubjeka: It is compatible with desire for coherence and modeling Consistency. This is the only place where there is rate limiting logic for the user api.

githubjeka: This is not an ideal implementation and just demo code for subsequent implementations.

src/User/Controller/RateLimitToUserApiMiddleware.php
```diff
@@ -0,0 +1,50 @@
...
```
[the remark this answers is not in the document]
githubjeka: done

config/web/rate-limit.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Yiisoft\Yii\RateLimiter\CounterInterface;
+use Yiisoft\Yii\RateLimiter\Storage\StorageInterface;
+use \Yiisoft\Yii\RateLimiter\Counter;
+
+/** @var array $params */
+
+return [
+    StorageInterface::class => function () {
+        $cache = new \Yiisoft\Cache\File\FileCache(__DIR__ . '/../runtime/rate-limit/');
```
xepozz, line 13: How about to use `Yiisoft\Aliases\Aliases` with `@runtime` value?
[Here](https://github.com/yiisoft/demo/blob/master/config/params.php#L52) you can see all aliases

config/routes.php
```diff
@@ -42,12 +46,17 @@
...
     // Auth
     Route::methods([Method::GET, Method::POST], '/login')
+        ->middleware(LimitRequestsMiddleware::class)
         ->action([AuthController::class, 'login'])
         ->name('auth/login'),
     Route::post('/logout')
         ->action([AuthController::class, 'logout'])
         ->name('auth/logout'),
     Route::methods([Method::GET, Method::POST], '/signup')
+        ->middleware(fn(\Psr\Container\ContainerInterface $container) => new LimitRequestsMiddleware(
```
xepozz, line 56: You can inject only needed objects without using the container. It's just the same, but a bit cleaner

xepozz, line 56:
```suggestion
        ->middleware(fn(ResponseFactoryInterface $responseFactory, StorageInterface $storage) => new LimitRequestsMiddleware(
```

config/web/rate-limit.php
githubjeka: [FileCache::__construct](https://github.com/yiisoft/cache-file/blob/master/src/FileCache.php#L105) does not support aliases

xepozz: And it should not.
[Here](https://github.com/yiisoft/demo/blob/de3b6fe1ae1d1a1429d5ed5f51aa467ec59a8714/config/common/translator.php#L15) is an example how to use `aliases`.
`$aliases->get('@runtime/rate-limiter')` === `/path/to/runtime/rate-limiter`

githubjeka: Thanks

samdark:
👍

Issue #147, opened 2020-10-01
how to add backend for the app?


TianNex:
start a new project? any suggestions?

### Additional info

| Q                | A
| ---------------- | ---
| Version          | 1.0.?
| PHP version      | 
| Operating system |

samdark:
What do you mean by "add backend"?

TianNex:
just like the backend of yii2-app-advanced

samdark:
In a similar way i.e. create another entry point etc. but there's almost no need for it in Yii 3 since it's way more flexible about directory structure.

jasonkovacs-tim:
@samdark Could you please describe how we might reorganize a large Yii2 advanced app (backend, frontend, console, and single vendor folder) into a directory structure that is recommended for Yii3?  I am reviewing https://github.com/yiisoft/app and this demo, but it is not clear what the migration path is from a Yii2 advanced app to Yii3 single app.

Would each of our top level apps sit side-by-side, each using the Yii3 single app directory structure, so that backend and frontend can each have config/ and a public/ entry point?  If that is done, would there need to be multiple sets of vendor/ and composer.json for each of our backend and frontend apps?  That seems like a lot of redundancy, though flexible.

Also, we rely heavily on the Yii2 concept of modules and routing with a tiered structure: app > modules > controllers.  I see in the demo that src/ has Blog/, and the config/routes.php has a Route Group of /blog and several Blog/*Controller entries named specifically.  In Yii3 Routing, is there a concept of pattern matching of a URI path part to a src/ subdirectory "module" and Controller without specifically naming them all in routes.php?  Could you provide an example of this pattern structure?  This is unclear from https://github.com/yiisoft/router.  Thank you.

rustamwin:
We recently added host support to the router, now you can split the admin (aka backend) by specifying the host. For example:
```php
Group::create()->routes(
    // your admin routes here.
)->host('admin.{_host}') // just use FastRoute regex
```

jasonkovacs-tim:
Will documentation be provided for how to manually migrate a former Yii2 advanced application template to a Yii3 single application?

samdark:
@jasonkovacs-tim eventually.

PR #339, opened 2021-08-24
feat: Started demo of backend app

githubjeka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #147

Availible on URL http://backend.localhost/app

image

### DIFF
--- a/config/routes-backend.php
+++ b/config/routes-backend.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+use Yiisoft\Router\Route;
+
+return [
+    Route::get('/app')
+        ->action([\App\Backend\Controllers\SiteController::class, 'index'])
+        ->name('backend/index')
+        ->host('backend.{_host}'),
+];
--- a/src/Backend/Controllers/SiteController.php
+++ b/src/Backend/Controllers/SiteController.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Backend\Controllers;
+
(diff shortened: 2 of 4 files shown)


config/routes-backend.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+use Yiisoft\Router\Route;
+
+return [
+    Route::get('/app')
```
rustamwin, line 8: I'd suggest using `Group` instead.
```php
Group::create('/')->routes(
    Route::get('')
        ->action([\App\Backend\Controllers\SiteController::class, 'index'])
        ->name('index'),
)->host('backend.{_host}')->namePrefix('backend/'),
```

src/Backend/views/site/index.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
rustamwin, line 1: View files in `src` isn't a good idea. What about `resources/backend/views` directory?

src/Backend/Controllers/SiteController.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Backend\Controllers;
+namespace App\Backend\Controller;
```
rustamwin, line 5: Directory name should be renamed as well.

githubjeka: :) magic github buttons...
will be done

PR #383, opened 2021-11-04
Fix scrutinizer-ci and code-coverage.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -34,7 +34,7 @@ build:
       tests:
         override:
           - command: |
-              php -S 127.0.0.1:8080 -t public public/index.php > ./runtime/yii.log 2>&1 &
+              ./yii serve 127.0.0.1:8080 > ./runtime/yii.log 2>&1 &
               vendor/bin/codecept run --coverage-xml
             on_node: 1
             coverage:
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -11,6 +11,9 @@ build:
   image: default-bionic
 
   environment:
+    variables:
+      YII_ENV: test
+
     php:
       version: 8.0.11
       ini:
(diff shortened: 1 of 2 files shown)


devanych (approved):

public/index.php
```diff
@@ -23,7 +23,7 @@
 
 require_once dirname(__DIR__) . '/preload.php';
 
-if ($_ENV['YII_ENV'] === 'test') {
+if ($_ENV['YII_ENV'] === 'test' || getenv('YII_ENV') === 'test') {
```
samdark, line 26: Should be no need for it since it's taken care of in https://github.com/yiisoft/demo/blob/master/preload.php#L12. If dotenv wasn't able to load it, that's quite bad but it should be fixed in preload anyway.

samdark: We should use `getenv('YII_DEV')`. See https://mattallan.me/posts/how-php-environment-variables-actually-work/

samdark:
```suggestion
if (getenv('YII_ENV') === 'test') {
```

.scrutinizer.yml
```diff
@@ -31,7 +34,7 @@ build:
       tests:
         override:
           - command: |
-              php -S 127.0.0.1:8080 -t public public/index-test.php > ./runtime/yii.log 2>&1 &
+              php -S 127.0.0.1:8080 -t public public/index.php > ./runtime/yii.log 2>&1 &
```
samdark, line 37:
```suggestion
              YII_ENV=test ./yii serve 127.0.0.1:8080 > ./runtime/yii.log 2>&1 &
```

samdark (approved):

PR #385, opened 2021-11-06
ApiInfo namespace change

wispoz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
Becase ApiInfo is Standalone actions , i think it must be in Controller\Actions namespace. If user whant add some actions he can place them in Actions folder. Its will be more clear to me.


samdark (approved):

samdark:
Agree. Thanks.

PR #392, opened 2021-11-13
Adopt router changes

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  | -

:warning: Merge after https://github.com/yiisoft/router/pull/128

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -24,6 +24,7 @@ use Yiisoft\DataResponse\Middleware\FormatDataResponseAsHtml;
 use Yiisoft\DataResponse\Middleware\FormatDataResponseAsJson;
 use Yiisoft\DataResponse\Middleware\FormatDataResponseAsXml;
 use Yiisoft\Http\Method;
+use Yiisoft\Router\CurrentRouteInterface;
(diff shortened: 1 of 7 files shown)


vjik (approved):

samdark (approved):

TianNex:
hello,is it really better than before ?
before ,it can get id form parameter $request directly,now we get it form additional parameter $currentRoute
: public function index(Request $request ): Response {

        $id = $request->getAttribute('id',null);

 
public function index(Request $request, CurrentRouteInterface $currentRoute): Response {

        $id = $currentRoute->getArgument('id',null);

vjik:
> hello,is it really better than before ?

Before that, there could be a conflict of attribute names (attributes with equal name from route and from something middleware).

samdark:
That's like getting your parameters from `$_REQUEST` instead of `$_GET`, `$_POST` in the old times. Uniform, easier to remember, may lead to serious issues.

PR #409, opened 2021-12-11
Add final to all classes.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Blog/Entity/Comment.php
+++ b/src/Blog/Entity/Comment.php
@@ -24,7 +24,7 @@ use DateTimeImmutable;
  *     }
  * )
  */
-class Comment
+final class Comment
 {
     /**
      * @Column(type="primary")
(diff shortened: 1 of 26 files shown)


src/Blog/Entity/Comment.php
```diff
@@ -24,7 +24,7 @@
  *     }
  * )
  */
-class Comment
+final class Comment
```
roxblnfk, line 27: It is bad idea to do finalized entity classes when you are using cycle orm

terabytesoftw: `Cycle/Orm` extends the identity classes, it's added in the docs, and I change it.

terabytesoftw: fixed.

samdark:
👍

Issue #418, opened 2021-12-24
Use Attributes instead of Annotations on entities


rustamwin:
There is no point in using annotations since we are using PHP 8.0

PR #421, opened 2021-12-28
Cycle orm v2

roxblnfk:
| Q             | A
| ------------- | ---
| New feature?  | ✔️

- Use attributes instead of annotations
- use ORM v2 instead of v1
  - cycle/database v2
  - DTO connection config
  - Behaviors instead of custom mappers

Some fixes


public/index.php
```diff
@@ -21,15 +21,16 @@
...
-    $c3 = dirname(__DIR__) . '/c3.php';
+    $c3 = 'c3.php';
     if (file_exists($c3)) {
         require_once $c3;
     }
 }
 
 // Run HTTP application runner
-$runner = new HttpApplicationRunner(dirname(__DIR__), $_ENV['YII_DEBUG'], $_ENV['YII_ENV']);
+$runner = new HttpApplicationRunner(getcwd(), $_ENV['YII_DEBUG'], $_ENV['YII_ENV']);
```
samdark, line 35: What is the reason for this change?

roxblnfk: try to compare current working dir in cases:
- `php yii serve`
- `php -S 127.0.0.1:8080 -t public` (like in [tests](https://github.com/yiisoft/demo/blob/15504bd4383685f2a095e857448d5c274315fab0/.github/workflows/build.yml#L65))

rustamwin:
#418

public/index.php
samdark: That won't work if you'll launch `./yii` not from the same directory as it is in:

```sh
./apps/yii3_demo/yii serve
```

roxblnfk: > That won't work

Feel free to fix it

samdark (approved):

samdark:
👍

PR #445, opened 2022-03-14
Environments

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌

### DIFF
--- a/config/params.php
+++ b/config/params.php
@@ -190,20 +190,20 @@ return [
          * ]
          */
         'schema-providers' => (Env::getBoolean('YII_DEBUG') ? [] : [
-                \Yiisoft\Yii\Cycle\Schema\Provider\SimpleCacheSchemaProvider::class => ['key' => 'cycle-orm-cache-key'],
-            ]) + [
-                // Store generated Schema in the file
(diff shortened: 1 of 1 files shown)


src/Env.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public static function getInteger(string $name, int $default = 0): int
+    {
+        self::ensureValuesLoaded();
+
+        return \filter_var(self::$values[$name] ?? null, FILTER_VALIDATE_INT, FILTER_NULL_ON_FAILURE) ?? $default;
+    }
+
+    public static function load(): void
```
vjik, line 37: Maybe move this method to separate class? In most cases this method will be unnecessary in IDE autocompletion on `Env::`.

samdark (approved):

rustamwin:
We need to discuss whether environment variables should be cached.

samdark:
What's the purpose of caching them?

BoShurik:
The advantage of the env variables that they can be changed on the fly. Without that you may use simple php array to store such variables

Issue #439, opened 2022-02-11
./yii fixture/add console command not working


arogachev:
According Codeception test fails as well.

### What steps will reproduce the problem?

Run `./yii fixture/add` console command`.

### What is the expected result?

No errors.

### What do you get instead?

```
./yii fixture/add
                                                                                                                        
 [ERROR] Transaction can't be finished. Some relations can't be resolved:                                               
Create new `postTag`                                                                                           
- tag.posts:postTag (Cycle\ORM\Relation\ShadowBelongsTo)  
```

### Additional info

| Q                | A
| ---------------- | ---
| Version          | current master
| PHP version      | https://github.com/yiisoft/demo/blob/3d6ca817d4efd08a4acf1a3a93521422c5764421/docker-compose.yml#L5
| Operating system | https://github.com/yiisoft/demo/blob/3d6ca817d4efd08a4acf1a3a93521422c5764421/docker-compose.yml#L5

rossaddison:
Encountering a similar issue with all my BelongsTo relations: 

Cycle\ORM\Exception\TransactionException: Transaction can't be finished. Some relations can't be resolved:
Update `product`
 - family (Cycle\ORM\Relation\BelongsTo) in C:\wamp64\www\yii-invoice\vendor\cycle\orm\src\Exception\TransactionException.php:40
Stack trace:
#0 C:\wamp64\www\yii-invoice\vendor\cycle\orm\src\Transaction\UnitOfWork.php(104): Cycle\ORM\Exception\TransactionException::unresolvedRelations(Array, Object(Cycle\ORM\Service\Implementation\RelationProvider))
#1 C:\wamp64\www\yii-invoice\vendor\cycle\orm\src\Transaction.php(48): Cycle\ORM\Transaction\UnitOfWork->run()
#2 C:\wamp64\www\yii-invoice\vendor\yiisoft\yii-cycle\src\Data\Writer\EntityWriter.php(30): Cycle\ORM\Transaction->run()
#3 C:\wamp64\www\yii-invoice\src\Invoice\Product\ProductRepository.php(59): Yiisoft\Yii\Cycle\Data\Writer\EntityWriter->write(Array)
#4 C:\wamp64\www\yii-invoice\src\Invoice\Product\ProductService.php(30): App\Invoice\Product\ProductRepository->save(Object(App\Invoice\Entity\Product Cycle ORM Proxy))
[...]

PR #449, opened 2022-04-08
Cycle ManyToMany relation is not working

albertborsos:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

I just want to inform you to know that the ManyToMany relation is not working correctly even in this demo application.
(Or I use it wrong, I can not solve this issue since days.)

I want to store multiple tags for posts. These tags maybe new, maybe already exists in the database. If it is new, then it works great. But if I want to assing an existing Tag to a new Post, it fails.

Related issue in Cycle ORM: https://github.com/cycle/orm/issues/329

I also updated the tests for this demo app, because I did not understand how it could work here. But it is not working.


albertborsos:
#439 is a related issue too

samdark:
@albertborsos how exactly does it fail?
@roxblnfk would you please take a look?

TiaNex-Com:
yes,

roxblnfk (commented):
I think it will be better to set ORM version to `^2.1.1`

PR #452, opened 2022-04-15
Fix errors

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | 

Uses https://github.com/yiisoft/validator/pull/222


src/Auth/Form/LoginForm.php
```diff
@@ -8,7 +8,7 @@
 use Yiisoft\Form\FormModel;
 use Yiisoft\Translator\TranslatorInterface;
 use Yiisoft\Validator\Result;
-use Yiisoft\Validator\Rule\Required;
+use Yiisoft\Validator\Rule\Required\Required;
```
samdark, line 11: Namespace looks weird. It was better before.

PR #454, opened 2022-05-17
Adopt to big refactoring of forms

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

Adopt to https://github.com/yiisoft/form/pull/192

### DIFF
--- a/config/params.php
+++ b/config/params.php
@@ -55,30 +55,36 @@ return [
         ],
     ],
 
-    'yiisoft/forms' => [
-        'field' => [
-            'ariaDescribedBy' => [true],
-            'containerClass' => ['form-floating mb-3'],
-            'errorClass' => ['fw-bold fst-italic invalid-feedback'],
-            'hintClass' => ['form-text'],
-            'inputClass' => ['form-control'],
-            'invalidClass' => ['is-invalid'],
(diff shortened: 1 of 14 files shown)


config/params.php
```diff
@@ -55,31 +55,25 @@
...
-            'defaultValues' => [
-                [
-                    'submit' => [
-                        'definitions' => [
-                            'class()' => ['btn btn-primary btn-lg mt-3'],
-                        ],
-                        'containerClass' => 'd-grid gap-2 form-floating',
+    'yiisoft/form' => [
+        'configs' => [
+            'default' => [
```
samdark, line 60: What's the benefit configuring form defaults instead of widgets?

samdark (approved):

config/params.php
vjik: > What's the benefit configuring form defaults instead of widgets?

This configuration of `FieldFactory`, that used for configure all fields widgets. For configure widgets via `yiisoft/widgets` we would have to set this paramateres for each widgets and for each new or custom widgets we also would also have to add this parameters.

Issue #458, opened 2022-06-08
BlogPageCest test fails


sankaest:
### What steps will reproduce the problem?

1.  composer install/update
2. ./yii fixture/add 20
3.  dev-tool/dev/demo$ vendor/bin/codecept run 

### What is the expected result?
✔ BlogPageCest: Blog page works.

### What do you get instead?
✖ BlogPageCest: Blog page works.
BlogPageCest test fails, if DB is filled with any sample data (eg, during istallation: ./yii fixture/add 20). 

### Additional info (from terminal)
There was 1 failure:
---------
1) BlogPageCest: Blog page works.
 Test  tests/Acceptance/BlogPageCest.php:testBlogPage
 Step  Am on page "/signup"
 Fail  Failed asserting that  on page /blog
-->  Yii Demo - Blog Yii Demo Blog Comments Feed Users Contact Swagger Language English Русский Login Signup Blog Showing 3 out of 22 posts Rem in et aut aut earum cumque. May, 27 by Federico6643And he got up and throw us, with the bread-knife.' The March Hare interrupted, yawning. 'I'm getting tired of
[Content too long to display. See complete response in '.../dev/demo/runtime/tests/_output/' directory]
--> contains "No records".

| Q                | A
| ---------------- | ---
| Version          | dev
| PHP version      | PHP 8.0.16 
| Operating system | Ubuntu 20.04

Need also more research, why we see failure:  Step  Am on page "/signup", despite in file BlogPageCest.php in function _before() we have: $I->amOnPage('/blog');

vjik:
https://github.com/yiisoft/demo/pull/461 don't fix problem. PR add command for clear tables in DB only.

sankaest:
What is the error? The same?

vjik:
> What is the error? The same?

yes.

PR #461, opened 2022-06-15
Clear db during testing

sankaest:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  | #458

### DIFF
--- a/src/Command/Fixture/SchemaClearCommand.php
+++ b/src/Command/Fixture/SchemaClearCommand.php
@@ -9,11 +9,11 @@ use App\Blog\Entity\Tag;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Yiisoft\Yii\Console\ExitCode;
 use Yiisoft\Yii\Cycle\Command\CycleDependencyProxy;
 use App\Blog\Entity\Comment;
 use App\Blog\Entity\Post;
 use App\User\User;
-use function count;
 
 
 final class SchemaClearCommand extends Command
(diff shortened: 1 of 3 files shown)


src/Command/Fixture/SchemaClearCommand.php
```diff
@@ -0,0 +1,95 @@
...
+                ->getRepository(Tag::class)
+                ->findAll())+
+            count($this->promise
+                ->getORM()
+                ->getRepository(User::class)
+                ->findAll())+
+            count($this->promise
+                ->getORM()
+                ->getRepository(Comment::class)
+                ->findAll());
```
BoShurik, line 91: findAll + count looks like huge overhead. Maybe replace it with count query?

src/Command/Fixture/SchemaClearCommand.php
```diff
@@ -0,0 +1,95 @@
...
+            ->delete('user')
+            ->run();
+
+        $this->promise
+            ->getDatabaseProvider()
+            ->database()
+            ->delete('comment')
+            ->run();
+
+        return count($this->promise
```
BoShurik, line 72: Command must return code. 0 - success, 1 - failure etc (see https://www.redhat.com/sysadmin/exit-codes-demystified)

BoShurik (changes requested):

src/Command/Fixture/SchemaClearCommand.php
sankaest: I was looking for that, but could not find it (I am new to Yii3). Will try to find it.

src/Command/Fixture/SchemaClearCommand.php
sankaest: Thank you for info!

samdark (approved):

samdark:
Thanks.

PR #476, opened 2022-07-18
Single approach to user creation

githubjeka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️

SignupForm has all rules for valid create a user.
The console comman `user/create` use SignupForm to create `User`
The web controller `SignupController` use SignupForm to create User


terabytesoftw:
I think it's best to move it to common, so we don't have duplicate settings.

src/Auth/Form/SignupForm.php
```diff
@@ -73,8 +73,12 @@ function (): Result {
                     $result->addError($this->translator->translate('validator.password.not.match'));
                 }
 
-                if ($result->getErrors() === [] && !$this->authService->signup($this->login, $this->password)) {
-                    $result->addError($this->translator->translate('validator.user.exist'));
+                if ($result->getErrors() === []) {
+                    try {
+                        $this->authService->signup($this->login, $this->password);
```
xepozz, line 78: Validation as creating a real user? Who made it? 😨

githubjeka: I didn't like it either. But the changes would affect other files. This PR would be huge. So I left it as is.

xepozz: I agree that it's better to do separately. I was just confused by this code and couldn't get past.

config/common/users.php
```diff
@@ -0,0 +1,19 @@
...
+use App\Auth\Identity;
+use App\Auth\IdentityRepository;
+use Cycle\ORM\ORMInterface;
+use Psr\Container\ContainerInterface;
+use Yiisoft\Auth\IdentityRepositoryInterface;
+
+/** @var array $params */
+
+return [
+    IdentityRepositoryInterface::class => static function (ContainerInterface $container): IdentityRepository {
```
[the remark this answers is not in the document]
githubjeka: I will fix these problems

githubjeka:
Sketch ready. Any comments?
I will fix config:
-  `config/common/users.php`
- tests

config/common/users.php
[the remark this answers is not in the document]
vjik: `Identity` is part of `User`, so this is good place for definition.

src/User/UserLogin.php
```diff
@@ -0,0 +1,51 @@
...
+        return $this->value;
+    }
+
+    /**
+     * @throw UserLoginException
+     */
+    public static function createNew(string $login, UserRepository $userRepository): UserLogin
+    {
+        $validator = new Validator(new SimpleRuleHandlerContainer());
+        $rules = [
```
vjik, line 30: Maybe move this validation into User ?

githubjeka: UserLogin can't be in invalid. The validation should belong to him. You can't create invalid a login. 

You could use it for many cases, for example type hint in public setter methods of User, forms, console args and others.
User is orm model and you could use a schema typecast handler.

src/User/UserPassword.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        return $this->hasher->validate($this->value, $hash);
+    }
+
+    /**
+     * @throw UserPasswordException
+     */
+    public static function createNew(string $password): UserPassword
+    {
+        $validator = new Validator(new SimpleRuleHandlerContainer());
```
samdark, line 39: Assetions isn't what validator was meant for. I think that should stay at input validation level.

src/User/SignupService.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\User;
+
+final class SignupService
+{
+    private UserLogin $login;
+    private UserPassword $password;
```
samdark, line 10: That looks wrong since we are having state here and services in general should not have state. These two look like they should be part of the singup() method i.e. `signup($login, $password)`.

src/User/UserLogin.php
samdark: Validation is not the same as asserting invariants. Validator wasn't meant for assertions. Also, checking for existing user is part of the business logic, not part of the invariant.

samdark (changes requested):

samdark (approved):

samdark:
👍

PR #483, opened 2022-07-29
Creating cookie _language for HTTP

githubjeka:
1. create host `127.0.0.1 kubernetes.docker.internal`
2. open http://kubernetes.docker.internal:8080/ru/ for RU language
3. Open http://kubernetes.docker.internal:8080/ 

### What is the expected result?
RU language

### What do you get instead?
default language EN

cookie _language missed
image

PS
Disabled Secure attribute of cookie for _language
https://web.dev/when-to-use-local-https/#use-http:localhost-by-default


samdark:
Thanks.

PR #491, opened 2022-08-04
Uploading files via Contact form

githubjeka:
Added missed multipart/form-data
Added missed multiple name attribute
Closing DIV tag matches nothing

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


rustamwin (approved):

arogachev (approved):

arogachev:
Tested it, the error 

```
`TypeError`
Cannot assign string to property `App\Contact\ContactForm::$attachFiles` of type `?array `
```

is gone.

But now the handling needs to be fixed too:

```
`Error`
Call to a member function `getStream()` on array 
```

PR #494, opened 2022-08-08
YiiSoft | Proposal to structure demo application

maximniko:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌

I took some time to form a more convenient application structure.
If you accept my offer, check the work of the tests.
Failed to quickly launch the application and tests as described in Readme.md

Я уделил немного времени для формирования более удобной структуры приложения.
Если принимаете мое предложение, проверьте работу тестов.
Не удалось быстро запустить приложение и тесты по описанию из Readme.md


config/events-console.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-use App\Timer;
+use App\Services\Timer;
```
roxblnfk, line 5: Don't use plural form in namespaces

xepozz (changes requested):
there is no structured modules

arogachev:
Could you briefly describe the difference?

samdark:
I agree with @xepozz. The proposed structure divides classes by-type rather than by use-case and that's not what we want to showcase. Structuring by use-case is preferred.

samdark:
Thanks for the effort but we'll stay with current demo structure and improve on it instead.

PR #501, opened 2022-09-05
Add yiisoft/request-model package #497

kv4nt:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #497


samdark:
I think @rustamwin meant actually using it, not only adding it to `composer.json`.

kv4nt:
Got it, ok.

PR #504, opened 2022-09-21
Add gii

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

- [x] Added Gii support
- [x] Raised PHP version up to 8.1


samdark:
Ready for review?

xepozz:
Yeap, but actually it doesn't work because of `config/params-web.php`. If I leave the settings from the `config/params-web.php` in the `params.php` it will break some `yii-gii` processes.

rustamwin:
FYI, I'll remove `view` dependency from Gii soon.

.github/workflows/build.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     name: PHP ${{ matrix.php }}-${{ matrix.os }}
 
     env:
-      extensions: fileinfo, pdo, pdo_sqlite, intl
+      extensions: fileinfo, pdo, pdo_sqlite, intl-70.1
```
samdark, line 34: Why 70.1?

tests/HttpClient/.gitignore
```diff
@@ -0,0 +1 @@
+*.private.env.json
```
samdark, line 1: Not sure we need that.

.github/workflows/build.yml
xepozz: Windows suddenly failed without it

samdark (approved):

PR #512, opened 2022-10-14
[Batch] Add composer-require-checker CI

xepozz:
Added composer-require-checker according to https://github.com/yiisoft/actions/pull/9

### DIFF
--- a/.github/workflows/composer-require-checker.yml
+++ b/.github/workflows/composer-require-checker.yml
@@ -1,0 +1,33 @@
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'CHANGELOG.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/composer-require-checker.yml
```diff
@@ -0,0 +1,33 @@
...
+name: Composer require checker
+
+jobs:
+  composer-require-checker:
+    uses: yiisoft/actions/.github/workflows/composer-require-checker.yml@master
+    with:
+      os: >-
+        ['ubuntu-latest']
+      php: >-
+        ['8.0', '8.1']
```
vjik, line 33:
```suggestion
        ['8.0']
```

vjik (changes requested):
Need adopt action to internal repositories

PR #515, opened 2022-10-18
fix: Timeout for ".\yii serve"

githubjeka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |  To fix The process ".\yii serve" exceeded the timeout of 300 seconds.

image

P.S. see https://getcomposer.org/doc/06-config.md#process-timeout

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -109,7 +109,10 @@
         }
     },
     "scripts": {
-        "serve": "./yii serve",
+        "serve": [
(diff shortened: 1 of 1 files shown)


arogachev (approved):
Thanks! :+1:

PR #517, opened 2022-10-18
Refactor code to PHP 8 style

githubjeka:


arogachev:
This was also fixed earlier as a part of #506 - https://github.com/yiisoft/demo/blob/c3cffaf48cc971921bb7317d38ae4b0ffc2445e5/src/Middleware/LocaleMiddleware.php. After adding rector such replacements will be done automatically. But until then, let's merge it separately. Thanks! :+1:

arogachev (approved):

PR #551, opened 2022-11-29
Ignore inspector requests

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌


xepozz (changes requested):
Should be ignore in debug-api packages. AFAIR it's already there

rustamwin:
> AFAIR it's already there

I think you are confused with yii-debug config

xepozz:
> > AFAIR it's already there
> 
> I think you are confused with yii-debug config

You are right.

xepozz (approved):

PR #576, opened 2023-03-12
Js panel example

xepozz:


blog/resources/debug/cache/modulefederation.config.js
```diff
@@ -0,0 +1,28 @@
+const { dependencies } = require('./package.json');
```
samdark, line 1: Should the cache be in git?

blog/resources/debug/cache/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "cra_remote",
```
samdark, line 2: cra?

blog/resources/debug/cache/modulefederation.config.js
xepozz: it's not cache, it's sources for cache panel

blog/resources/debug/cache/package.json
xepozz: **C**reate **R**eact **A**pp

PR #579, opened 2023-03-27
Attribute routes example

xepozz:
image
image
image

In development I just run the worker in console `php watcher.php` and it rebuild the attributes map as fast as possible:
image

Related: https://github.com/yiisoft/router/pull/194

### DIFF
--- a/blog/config/common/routes/routes.php
+++ b/blog/config/common/routes/routes.php
@@ -17,6 +17,7 @@ use App\Middleware\AccessChecker;
 use App\Middleware\ApiDataWrapper;
 use App\User\Controller\ApiUserController;
 use App\User\Controller\UserController;
+use olvlvl\ComposerAttributeCollector\Attributes;
 use Psr\Http\Message\ResponseFactoryInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Yiisoft\Auth\Middleware\Authentication;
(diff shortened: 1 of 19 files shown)


samdark:
Nice. That should at least go into cookbook.

blog/watcher.php
```diff
@@ -0,0 +1,22 @@
...
+
+`php watcher-build.php`;
+
+echo 'Ready for listening changes...' . PHP_EOL;
+
+Watch::paths(__DIR__ . '/src')
+    ->onAnyChange(function (string $type, string $path) {
+        echo sprintf('File changed: "%s".', $path) . PHP_EOL;
+        echo 'Run watcher-build...' . PHP_EOL;
+        `php watcher-build.php`;
```
rustamwin, line 19: Isn't it disabled by default?

blog/config/common/routes/routes.php
```diff
@@ -36,11 +37,22 @@
 use Yiisoft\Yii\RateLimiter\LimitRequestsMiddleware;
 use Yiisoft\Yii\RateLimiter\Storage\StorageInterface;
 
+$routes = [];
+foreach (Attributes::findTargetMethods(Route::class) as $method) {
```
rustamwin, line 41: What about classes and groups?

blog/watcher.php
xepozz: Disabled what?

blog/config/common/routes/routes.php
xepozz: It's just an example. If we want to merge it I'll migrate all routes and add groups and classes

